PR #109, opened 2021-09-16
addpkg: libretro-kronos

IEAST:


libretro-kronos/fix-unrecognized-msse.patch
```diff
@@ -0,0 +1,14 @@
...
+--- a/yabause/src/libretro/Makefile	2021-09-16 08:58:50.536977235 +0000
++++ b/yabause/src/libretro/Makefile	2021-09-16 09:00:29.773688502 +0000
+@@ -261,10 +261,6 @@
+ 	FLAGS += -D__USE_OPENGL_DEBUG__
+ endif
+ 
+-ifeq ($(HAVE_SSE),1)
+-	FLAGS += -msse -mfpmath=sse
+-endif
+-
```
felixonmars, line 11: You should probably find out why `HAVE_SSE` is 1 here, rather than deleting it altogether.

felixonmars (changes requested):

PR #143, opened 2021-09-28
addpkg: pacparser

r-value:


felixonmars:
Thanks for your contribution.

Unfortunately we do not want to include fixes that should be applied to Arch upstream when it's not critical to RISC-V porting/packaging. The package in question actually builds fine now unmodified, probably because an dependency was upgraded.

Sorry for wasting your time here, but you may probably want to suggest these changes to Arch upstream instead.

PR #194, opened 2021-10-05
addpkg: upx

Xeonacid:
Newer commits of the submodule are all format changes, so I checkout the newest directly.


Xeonacid:
Closed since it builds fine now.

PR #227, opened 2021-10-10
addpkg: hatari

r-value:
This patch is ugly. `REG_A0` is already defined in `sys/ucontext.h` on RISC-V.

Request for better workaround.


r-value:
Better workaround is applied and ready to merge.

This patch may be upstreamed soon.

PR #252, opened 2021-10-19
addpkg: bsd-games

rapiz1:


bastianplettner:
`bsd-games` builds fine without any changes to Arch's PKGBUILD.
This patch is not needed.

PR #255, opened 2021-10-20
addpkg: python-autobahn

Ast-x64:


yan12125:
Hmm, strange. It looks like `-march=rv64gc` is specified by [makepkg](https://github.com/felixonmars/archriscv-packages/blob/f800753a24499bcb0f1b69a8728d893ab94d05f6/devtools/makepkg-riscv64.conf#L41), but not used by autobahn. Let me try to reproduce the issue with qemu.

PR #268, opened 2021-10-22
addpkg: surge

Ast-x64:


XieJiSS:
The `fix_pthread.patch` hack is not needed anymore.

Xeonacid:
This package has been removed in Arch. Closing.

PR #272, opened 2021-10-23
addpkg: rasqal

yujqiao:


felixonmars:
I have also reported this to Arch some time ago: https://bugs.archlinux.org/task/72101

PR #278, opened 2021-10-25
repair gdmap

snowpiaoling:
repair gdmap


r-value:
Closing this because the package `gdmap` is no longer maintained by Arch Linux.

PR #308, opened 2021-11-02
Clean GHC bootstrapping

hexchain:
Patches come from this tree: https://gitlab.haskell.org/hexchain/ghc/-/commits/ghc-9.0.1-rv64-llvm-bootstrap/

Patch files are generated with:
```
git format-patch --zero-commit --no-numbered ghc-9.0.1-release..ghc-9.0.1-rv64-llvm-bootstrap
```
and then treated with:
```
for f in *.patch; do filterdiff -x '*/.gitlab-ci.yml' -x '*/testsuite/*' $f | sponge $f; done
```
in order to remove hunks of files not present in release tarballs.


ghc/riscv64.patch
```diff
@@ -1,69 +1,77 @@
...
  
    cp mk/build.mk{.sample,}
--  sed -i '1iBuildFlavour = perf' mk/build.mk
-+  sed -i '1iBuildFlavour = perf-llvm' mk/build.mk
-+
-+  ./boot
- }
+   sed -i '1iBuildFlavour = perf-llvm' mk/build.mk
++  cat >>mk/build.mk <<EOF
++CONF_CC_OPTS_STAGE0 += -Wno-unused-label -Wno-unused-but-set-variable -Wno-unused-variable
```
hexchain, line 73: This is only meant to suppress excessive compiler warnings when building with an unregisterised compiler (i.e. `ghc-bootstrap`). It is not required at all in subsequent builds.

It could be a little bit cleaner if we just enable LLVM backend for unregisterised compilers with a one-line patch.

PR #321, opened 2021-11-05
addpkg: qalculate-qt

Ast-x64:


qalculate-qt/fix_arch.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/.qmake.stash b/.qmake.stash
```
felixonmars, line 1: Perhaps this file should not exist at all in the source tarball?

felixonmars, line 1: Perhaps this file should not exist at all in the source tarball?

moodyhunter:
Hello, @Ast-x64 why are you attempting to remove such a file in `prepare()` while this file doesn't exist at all from the release tarball?

r-value:
Hi @moodyhunter , this file do exist when this PR was merged. It disappears now and we'll remove this patch later

moodyhunter:
Thanks for your response

PR #333, opened 2021-11-11
addpkg: libretro-melonds

Ast-x64:


Xeonacid:
Confirmed packaged successfully without patch now. Closed.

PR #336, opened 2021-11-11
addpkg: libantlr3c

Ast-x64:


r-value (approved):
The `--disable-abiflags` means "Does not add ABI flags -m32 or -m64", which is what we expected.

PR #342, opened 2021-11-11
addpkg: gitlab-workhorse

Ast-x64:


XieJiSS:
I suspect that this is outdated because the upstream has updated `prometheus/procfs` to 0.9.0: https://gitlab.com/gitlab-org/gitlab/-/blob/master/workhorse/go.mod#L99

Xeonacid:
This package has been removed in Arch. Closing.

PR #357, opened 2021-11-13
addpkg: diskus

xctan:


diskus/riscv64.patch
```diff
@@ -0,0 +1,29 @@
...
++source=("diskus-$pkgver.tar.gz::https://static.crates.io/crates/diskus/diskus-$pkgver.crate"
++        fix-libc-version.patch)
++sha512sums=('c532de84356f43109242805b47f6154b0f837c176e5aa016c5f2c221ff2b77da91511cc6e01351924f38cffd5f26c7e8740a4475def1f7cbcffc29501c3321b1'
++            '8e1db6ebdadf5a028d3fb81ecc76ee4fa190bf23b86e80fd794de3b3f9353997b044181424c4adcff07fcf9d3d364d455891f64915a44cf6c7299527b16056f9')
+ 
+ prepare() {
+   cd diskus-$pkgver
+-  cargo fetch --locked
++  patch -Np1 -i ../fix-libc-version.patch
++  cargo fetch
```
r-value, line 20: > Please do not just remove `--locked`. Consider `cargo update -p <pkg_name>` instead to update only the culprit.

Ref: https://github.com/felixonmars/archriscv-packages/pull/248#discussion_r730910645

diskus/riscv64.patch
```diff
@@ -0,0 +1,29 @@
...
+ }
+ 
+ build() {
+   cd diskus-$pkgver
+-  cargo build --release --locked --offline
++  cargo build --release --offline
```
r-value, line 26: Same as above.

r-value (changes requested):

PR #358, opened 2021-11-15
addpkg: merkaartor

Xeonacid:


r-value:
It might be better to submit a proposal for Arch Linux to change build system from qmake to cmake.

moodyhunter:
@XieJiSS I don't know wtf happened to my blog post, but the updated links are

https://mooody.me/posts/2022-04/archrv-qmake6/ and 
https://mooody.me/posts/2022-04/qprocess-in-qemu/

PR #360, opened 2021-11-15
addpkg: hex

r-value:


felixonmars:
Please also send a PR to update the lock file upstream :)

r-value:
> Please also send a PR to update the lock file upstream :)

It seems that the [source repo](https://github.com/KokaKiwi/rust-hex) doesn't have a `Cargo.lock` file >_< so I just did nothing

felixonmars:
> > Please also send a PR to update the lock file upstream :)
> 
> It seems that the [source repo](https://github.com/KokaKiwi/rust-hex) doesn't have a `Cargo.lock` file >_< so I just did nothing

The source repo should be: https://github.com/sitkevij/hex

PR #376, opened 2021-11-17
addpkg: nomad

Ast-x64:


nomad/riscv64.patch
```diff
@@ -0,0 +1,42 @@
...
++        'b82609e4d183d8b82f693a84be388b150c6669326c4d4e43174da4b2f702e4686ae46a13f5a22444ae961dcb5049c31386223df0163997d628b2924d6dff9cc5')
+ 
+ prepare() {
+ 	cd "$pkgname-$pkgver"
+@@ -43,6 +46,8 @@ prepare() {
+ 
+   # download dependencies
+   go mod download
++  go mod vendor
++  patch -Np1 -i ../add_riscv_support.patch
```
r-value, line 31: I suggest change the vendor to an active fork, and send the vendor change to upstream
```suggestion
+  go mod edit -replace github.com/kr/pty=github.com/creack/pty
+  go mod edit -replace github.com/boltdb/bolt=github.com/etcd-io/bbolt
   go mod download
```

r-value (changes requested):
The patch for archived vendors may remain in the repo for a long time. I think it's not a good idea.

PR #383, opened 2021-11-19
addpkg: argocd

Ast-x64:


SpriteOvO:
This package can now be built without a patch.

PR #384, opened 2021-11-19
addpkg: libredefender

Ast-x64:


libredefender/riscv64.patch
```diff
@@ -0,0 +1,31 @@
...
++        'SKIP'
++        'cf82f590fcdf3b98c62a8494e685d8e1e53a7009e5361be91e8c480a55af7040dbbdf2ede22f49a5cbb027be37b75e225a7da37a4110d2c4d3165bb51ef75ddd'
++        '9226d72280a7b408575cb6789193ed61f1861e5c465510a7b1bf555e238f482e73b645d50d3b5d03dd6d7afe96fb975ba73ef6fd5accca3df0d2204c0ea75a6d')
+ 
+ validpgpkeys=("64B13F7117D6E07D661BBCE0FE763A64F5E54FD6")
+ 
+ prepare() {
+   cd "${pkgname}-${pkgver}"
+-  cargo fetch --locked --target "$CARCH-unknown-linux-gnu"
++  cargo fetch --locked --target "riscv64gc-unknown-linux-gnu"
```
r-value, line 24: I don't think it's a good idea to substitute the variables with specific values, which reduces the generality of `PKGBUILD`.

libredefender/riscv64.patch
```diff
@@ -0,0 +1,31 @@
...
++  cd ~/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.17.0/src/sys/ptrace
++  patch -Np1 -i "${srcdir}/fix_ptrace_enum.patch"
```
r-value, line 26: `nix-0.23.0`is available and compatible with RISC-V, please remove the patch and bump the dependency instead.

libredefender/fix_type.patch
```diff
@@ -0,0 +1,40 @@
+diff --git a/engine.rs b/engine.rs
+index 5f2fe24..e26bc58 100644
+--- a/engine.rs
++++ b/engine.rs
+@@ -78,7 +78,7 @@ pub struct Engine {
+ unsafe impl Send for Engine {}
+ unsafe impl Sync for Engine {}
+ 
+-fn map_scan_result(result: cl_error_t, virname: *const i8) -> Result<ScanResult, ClamError> {
++fn map_scan_result(result: cl_error_t, virname: *const u8) -> Result<ScanResult, ClamError> {
```
r-value, line 10: Are these changes compatible with x86_64? I really don't think that it's the source of the problem, could you please explain the reasons for this patch further?

r-value (changes requested):

r-value:
Any updates?

Xeonacid:
Superseded by #2895 , closing.

PR #385, opened 2021-11-19
addpkg: ipvsadm

Ast-x64:


r-value (approved):
Could get upstreamed to Arch Linux.

PR #386, opened 2021-11-19
addpkg: libidn11

Ast-x64:


r-value (approved):
This could probably be upstreamed to Arch Linux. Could you please also test it on x86_64?

PR #389, opened 2021-11-21
updpkg: linux

ZenithalHourlyRate:
This has been tested on unmatched.

Procedures to produce this `config` from oldconfig

```
make olddefconfig
# Manually edit .config to enable all SIFIVE related options like SOC, ERRATA, DRIVERS and espcially PCIE_FU740
make olddefconfig
```

Some note: I tested this kernel with the following initramfs as my rootfs is ext4 and on mmc
```
dracut --kver 5.15.3-arch1-1 --force-drivers "ext4 mmc_block mmc_spi pwrseq_emmc pwrseq_simple roles spi-sifive pwm-sifive" -f
```


ZenithalHourlyRate:
> Some note: I tested this kernel with the following initramfs as my rootfs is ext4 and on mmc

After ablation we found only `spi-sifive` is necessary for detecting rootfs. Dracut would load `ext4`, `mmc` automatically.

ZenithalHourlyRate:
> After ablation we found only `spi-sifive` is necessary

Further update on this: since dracut 56 all drivers in spi/ folder would be included in initrd if the architecture is riscv automatically, so this is not necessary now.

raunakwete43 (approved):

PR #395, opened 2021-11-24
addpkg: spice-gtk

r-value:
Could get upstreamed to Arch Linux.

Dependency updated according to error logs.

### DIFF
--- a/spice-gtk/riscv64.patch
+++ b/spice-gtk/riscv64.patch
@@ -1,0 +1,23 @@
+diff --git PKGBUILD PKGBUILD
+index 778bbc3b..9f64e23e 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -12,7 +12,7 @@ arch=('x86_64')
+ url="https://www.spice-space.org/"
+ license=('LGPL2.1')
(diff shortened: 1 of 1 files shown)


spice-gtk/riscv64.patch
```diff
@@ -0,0 +1,23 @@
...
+ provides=("spice-glib=$pkgver" "spice-gtk3=$pkgver")
+ replaces=('spice-glib' 'spice-gtk3')
+ source=("https://www.spice-space.org/download/gtk/$pkgname-$pkgver.tar.xz"{,.sig})
+@@ -22,7 +22,8 @@ sha256sums=('23acbee197eaaec9bce6e6bfd885bd8f79708332639243ff04833020865713cd'
+ validpgpkeys=('206D3B352F566F3B0E6572E997D9123DE37A484F') # Victor Toso de Carvalho <me@victortoso.com>
+ 
+ build() {
+-  arch-meson $pkgname-$pkgver build -D celt051=disabled
++  # CELT codec is disabled by default since v0.35
++  arch-meson $pkgname-$pkgver build
```
felixonmars, line 20: This part should be filed as a bug report to Arch, as an improvement. I believe it isn't specific for riscv64.

Xeonacid:
This package can be built successfully now without any patch.

PR #400, opened 2021-11-24
addpkg: mapnik

Ast-x64:


Ast-x64:
Seems this has been fixed by the upstream already.

PR #407, opened 2021-11-25
addpkg: rocksdb

XieJiSS:


XieJiSS:
Patch has been submitted to https://github.com/facebook/rocksdb/pull/9215 , but we have to keep a local patch because `9215.patch` can't be directly applied to `version=2.6.10`

PR #429, opened 2021-11-29
addpkg: openblas

rapiz1:


openblas/riscv64.patch
```diff
@@ -0,0 +1,29 @@
...
+ pkgrel=1
+ pkgdesc="An optimized BLAS library based on GotoBLAS2 1.13 BSD"
+ url="https://www.openblas.net/"
+ license=('BSD')
+ depends=('gcc-libs')
+@@ -20,9 +20,16 @@
+ build() {
+   cd $_pkgname-$pkgver
+ 
+-  make NO_STATIC=1 NO_LAPACK=1 NO_LAPACKE=1 NO_CBLAS=1 NO_AFFINITY=1 USE_OPENMP=1 \
```
felixonmars, line 14: We can skip the `$CARCH` branches for now, because patching only the TARGET would actually make the change more clear. As you can see below.

openblas/riscv64.patch
```diff
@@ -0,0 +1,29 @@
...
+-       CFLAGS="$CPPFLAGS $CFLAGS" TARGET=CORE2 DYNAMIC_ARCH=1 \
+-       NUM_THREADS=64 MAJOR_VERSION=3 libs shared
++  if [[ $CARCH -eq riscv64 ]]
++  then
++    make NO_STATIC=1 NO_LAPACK=1 NO_LAPACKE=1 NO_CBLAS=1 NO_AFFINITY=1 USE_OPENMP=1 \
++         CFLAGS="$CPPFLAGS $CFLAGS" TARGET=CORE2 DYNAMIC_ARCH=1 \
++         NUM_THREADS=64 MAJOR_VERSION=3 libs shared TARGET=RISCV64_GENERIC
```
felixonmars, line 21: You actually specified `TARGET` twice here :)

felixonmars (changes requested):

PR #438, opened 2021-11-30
addpkg: amdvlk

Ast-x64:


r-value (approved):
Upstreamed in https://github.com/GPUOpen-Drivers/pal/pull/85

PR #449, opened 2021-12-05
updpkg: devtools

IEAST:
updpkg to 20211129

### DIFF
--- a/devtools/riscv64.patch
+++ b/devtools/riscv64.patch
@@ -1,59 +1,42 @@
-diff --git PKGBUILD.orig PKGBUILD
-index 1046d6a..0e9938c 100644
---- PKGBUILD.orig
+diff --git PKGBUILD PKGBUILD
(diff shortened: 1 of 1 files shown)


devtools/riscv64.patch
```diff
@@ -1,59 +1,42 @@
...
+--- PKGBUILD
 +++ PKGBUILD
-@@ -3,9 +3,9 @@
- 
- pkgname=devtools
- pkgver=20210202
--pkgrel=3
-+pkgrel=3.1
- pkgdesc='Tools for Arch Linux package maintainers'
--arch=('any')
```
felixonmars, line 12: Actually, please don't remove this part. This is intended because the package differs in a qemu-enabled x86 builder (for x86_64) and in the riscv64 repository, so it's not "any".

IEAST: Ok. i will fix it

PR #497, opened 2021-12-22
addpkg: x42-plugins

Xeonacid:


x42-plugins/riscv64.patch
```diff
@@ -0,0 +1,32 @@
...
+index f246ee5d..7147ffc6 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -19,20 +19,24 @@ optdepends=('jack: for standalone applications'
+ source=("http://gareus.org/misc/${pkgname}/${pkgname}-${pkgver}.tar.xz"
+         "${pkgname}.png::https://x42-plugins.com/x42/static/img/x42.png"
+         "${pkgname}.directory"
+-        "${pkgname}.menu")
++        "${pkgname}.menu"
++        "fix-compile.patch")
```
felixonmars, line 11: Please use a more descriptive name for the patch, for example `disable-sse`

felixonmars (changes requested):

x42-plugins/riscv64.patch
Xeonacid: Fixed.

PR #560, opened 2022-01-10
addpkg: portmidi

cubercsl:
upstream's commit https://github.com/mixxxdj/portmidi/commit/7c82f8873962045d2aa6b1e02e1431c073ec13b4 may fix build with `-Werror-format-security`, but without a tag release.

### DIFF
--- a/portmidi/0001-fix-build-with-Werror-format-security.patch
+++ b/portmidi/0001-fix-build-with-Werror-format-security.patch
@@ -20,7 +20,7 @@ index 507648c..a3d5b4c 100755
      char line[STRING_MAX];
      int n = 0, i;
 -    printf(prompt);
-+    fputs(prompt, stdout);
++    printf("%s", prompt);
      while (n != 1) {
          n = scanf("%d", &i);
          fgets(line, STRING_MAX, stdin);
@@ -33,7 +33,7 @@ index 60fcf7a..4031ab0 100644
      char line[STRING_MAX];
      int n = 0, i;
 -    printf(prompt);
-+    fputs(prompt, stdout);
++    printf("%s", prompt);
      while (n != 1) {
          n = scanf("%d", &i);
          fgets(line, STRING_MAX, stdin);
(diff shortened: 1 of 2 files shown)


portmidi/0001-fix-build-with-Werror-format-security.patch
```diff
@@ -0,0 +1,117 @@
...
++    fputs(prompt, stdout);
+     while (n != 1) {
+         n = scanf("%d", &i);
+         fgets(line, STRING_MAX, stdin);
+@@ -256,7 +256,7 @@ int main(int argc, char **argv)
+     err = Pm_OpenOutput(&midi, outp, DRIVER_INFO, OUTPUT_BUFFER_SIZE, 
+                         TIME_PROC, TIME_INFO, LATENCY);
+     if (err) {
+-        printf(Pm_GetErrorText(err));
++        puts(Pm_GetErrorText(err));
```
r-value, line 45: `puts` will add additional `\n` at the end of the string. Applying only this patch will cause inconsistency. There might be other changes in `Pm_GetErrorText` which should be patched together.

portmidi/0001-fix-build-with-Werror-format-security.patch
```diff
@@ -0,0 +1,117 @@
...
++        else            puts(Pm_GetErrorText(count));
+     }
+ }
+ 
+@@ -168,7 +168,7 @@ int main(int argc, char **argv)
+     inp = get_number("Type input device number: ");
+     err = Pm_OpenInput(&midi_in, inp, NULL, 512, NULL, NULL);
+     if (err) {
+-        printf(Pm_GetErrorText(err));
++        puts(Pm_GetErrorText(err));
```
r-value, line 76: Same as above

portmidi/0001-fix-build-with-Werror-format-security.patch
```diff
@@ -0,0 +1,117 @@
...
++    fputs(prompt, stdout);
+     while (n != 1) {
+         n = scanf("%d", &i);
+         fgets(line, STRING_MAX, stdin);
+@@ -136,7 +136,7 @@ void receive_poll(PtTimestamp timestamp, void *userData)
+     if (!active) return;
+     while ((count = Pm_Read(midi_in, &event, 1))) {
+         if (count == 1) output(event.message);
+-        else            printf(Pm_GetErrorText(count));
++        else            puts(Pm_GetErrorText(count));
```
r-value, line 67: Same as above

a-fruity-melon:
@cubercsl

PR #565, opened 2022-01-11
updpkg: qt5-base

r-value:
Note that removal of `libfbclient` from `makedepends` might be unnecessary. Currently working on building `libfbclient` on RISC-V.


r-value:
Update: confirmed that `libfbclient` is built successfully on unmatched and changes about `libfbclient` is unnecessary.

PR #608, opened 2022-01-22
updpkg: rust

r-value:
`rust-1.56.0`:

 * Update to 1.56.0 (critical version for edition2021)
 * LLVM13 rebuild

This patch is based on following `PKGBUILD` file:
https://github.com/archlinux/svntogit-packages/blob/1398be71b3b05a8be83cede23a64bdf2d0d1f99e/trunk/PKGBUILD

Note:
 * This patch requires `llvm12-libs`
 * `makedepends` won't contain `llvm12-libs` in 1.58.0 update

### DIFF
--- a/rust/riscv64.patch
+++ b/rust/riscv64.patch
@@ -1,104 +1,120 @@
 diff --git PKGBUILD PKGBUILD
-index 522e5124..831c4254 100644
+index f05d522..5ec7c48 100644
 --- PKGBUILD
 +++ PKGBUILD
-@@ -4,7 +4,7 @@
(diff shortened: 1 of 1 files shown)


r-value:
updpkg: rust 1.57.0

This patch is based on the following `PKGBUILD` file:
https://github.com/archlinux/svntogit-packages/blob/f5349eee12ea36a8125a4b09f433600729b597d1/trunk/PKGBUILD

Note:
 * Option `lto` will cause some link failures
 * Feature `jemalloc` is fixed in unreleased changes. Enable it later.

PR #612, opened 2022-01-23
rmpkg: fd

rapiz1:
It has been upstreamed so there's no need to keep a patch
`v8.3.1` should be compiled once Rust has been fixed.

https://github.com/sharkdp/fd/commit/e5145ffb985ab1fdd72b680806dff4fc00c1877f


r-value:
Note: package `fd` didn't pass the checks on `qemu-user`.

```plain
failures:

---- test_opposing::uu stdout ----
thread 'test_opposing::uu' panicked at 'assertion failed: `(left == right)`
  left: `Output { status: ExitStatus(ExitStatus(0)), stdout: "./a.foo\n./e1 e2\n./one\n./one/b.foo\n./one/two\n./one/two/C.Foo2\n./one/two/c.foo\n./one/two/three\n./one/two/three/d.foo\n./one/two/three/directory_foo\n./symlink\n", stderr: "" }`,
 right: `Output { status: ExitStatus(ExitStatus(0)), stdout: "./symlink\n./one\n./a.foo\n./e1 e2\n./one/b.foo\n./one/two\n./one/two/C.Foo2\n./one/two/c.foo\n./one/two/three\n./one/two/three/d.foo\n./one/two/three/directory_foo\n", stderr: "" }`: --ignore --no-hidden should override -uu', tests/tests.rs:1976:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_opposing::uu

test result: FAILED. 72 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.57s
```

PR #520, opened 2021-12-24
upgpkg: mkinitcpio-busybox

hexchain:

PR #616, opened 2022-01-23
updpkg: mkinitcpio-busybox

rapiz1:


rapiz1:
Dup https://github.com/felixonmars/archriscv-packages/pull/520

PR #631, opened 2022-01-24
addpkg: python-socketio

Ast-x64:


cubercsl:
Probably due to python-engineio not up to date.

PR #644, opened 2022-01-26
addpkg: python-nose-progressive

Xeonacid:


Xeonacid:
The python-nose-progressive package is now in the AUR

PR #650, opened 2022-01-28
addpkg: ethtool

Avimitin:


ethtool/riscv64.patch
```diff
@@ -0,0 +1,13 @@
+diff --git PKGBUILD PKGBUILD
+index 55e891c..326ce8d 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -19,7 +19,7 @@ validpgpkeys=(CE4A4D080F0D304F23B9EBDD972D5BF4DC613806  # John W. Linville <linv
+ 
+ build() {
+     cd $pkgname-$pkgver
+-    ./configure --prefix=/usr --mandir=/usr/share/man --sbindir=/usr/bin
++    ./configure --prefix=/usr --mandir=/usr/share/man --sbindir=/usr/bin --enable-netlink=no
```
felixonmars, line 10: IIRC it was only a test failure. Are you sure the functionality isn't available at all?

Avimitin: I will try to figure it out.

Avimitin:
Ethtool can be built successfully without any modification if not in the QEMU.

PR #656, opened 2022-01-29
upgpkg: devtools 20220126-1.1

Xeonacid:


devtools/riscv64.patch
```diff
@@ -1,54 +1,46 @@
...
+ url='https://gitlab.archlinux.org/archlinux/devtools'
  depends=('bash' 'openssh' 'subversion' 'rsync' 'arch-install-scripts'
-@@ -12,7 +12,9 @@
-          'git' 'bzr' 'mercurial' 'diffutils' 'util-linux' 'awk')
+@@ -13,7 +13,8 @@
  makedepends=('asciidoc')
  optdepends=('btrfs-progs: btrfs support')
--source=(https://sources.archlinux.org/other/${pkgname}/${pkgname}-${pkgver}.tar.gz{,.sig})
-+source=(https://sources.archlinux.org/other/${pkgname}/${pkgname}-${pkgver}.tar.gz{,.sig}
-+        https://github.com/archlinux/devtools/pull/64.patch
```
felixonmars, line 23: Thanks for the update. As we are no longer patching the devtools code now, we should be splitting the additional files to a new package like `devtools-riscv64` so that we are no longer blocked by new devtools updates.

felixonmars (changes requested):

PR #701, opened 2022-02-04
addpkg: alot

XieJiSS:
Requires dependencies to be rebuilt before merge: `outdated_dep [python-magic] [python-urwidtrees]`


XieJiSS:
Upstream has a patch set https://www.mail-archive.com/notmuch@notmuchmail.org/msg50558.html but that won't fix this issue.

Since we have no access to `notmuch-0.34.3/lib/open.cc`, this is the only possible way to make the failing test succeed.

PR #708, opened 2022-02-04
addpkg: python-flask-dance

XieJiSS:
`python-flask-dance` > `python-sqlalchemy-utils` (check) > `python-cryptography` > `python-setuptools-rust`


Xeonacid:
This should be fixed in `python-cryptography` instead?

PR #717, opened 2022-02-05
addpkg: wakatime

piggynl:


wakatime/riscv64.patch
```diff
@@ -0,0 +1,33 @@
...
+ checkdepends=('svn')
+-source=("${pkgname}-cli::git+https://github.com/${pkgname}/${pkgname}-cli#tag=v${pkgver}")
+-sha256sums=('SKIP')
++source=("${pkgname}-cli::git+https://github.com/${pkgname}/${pkgname}-cli#tag=v${pkgver}"
++        "main_test-arch.patch::https://github.com/wakatime/wakatime-cli/pull/642.patch")
++sha256sums=('SKIP'
++            'eeea51959573aba7091357fa428fb60a90cc4148b788cc5d8d12f4adb4753461')
+ 
+-_binname="${pkgname}-cli-linux-amd64"
++_binname="${pkgname}-cli-linux-${CARCH}"
```
Xeonacid, line 15: This won't work for x86_64, so I think directly patching it to `riscv64` may be more clear.

cubercsl:
Need to be useful after https://github.com/wakatime/wakatime-cli/pull/642 is merged. Otherwise the test will fail.

r-value (changes requested):
Can't reproduce. Build fails in `check()`.
Error log:
```plain
==> Starting check()...
go: downloading github.com/stretchr/testify v1.7.0
go: downloading github.com/davecgh/go-spew v1.1.1
go: downloading github.com/pmezard/go-difflib v1.0.0
go: downloading gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
./build/wakatime-cli-linux-riscv64 --log-file /tmp/TestSendHeartbeats2731894363/002/1416661128 --api-url http://127.0.0.1:42365 --key 00000000-0000-4000-8000-000000000000 --config /tmp/TestSendHeartbeats2731894363/001/wakatime.cfg2941296879 --entity testdata/main.go --cursorpos 12 --offline-queue-file /tmp/TestSendHeartbeats2731894363/001/3667903834 --lineno 42 --lines-in-file 100 --time 1585598059 --hide-branch-names .* --project wakatime-cli --write --verbose


failed to run command ./build/wakatime-cli-linux-riscv64 --log-file /tmp/TestSendHeartbeats2731894363/002/1416661128 --api-url http://127.0.0.1:42365 --key 00000000-0000-4000-8000-000000000000 --config /tmp/TestSendHeartbeats2731894363/001/wakatime.cfg2941296879 --entity testdata/main.go --cursorpos 12 --offline-queue-file /tmp/TestSendHeartbeats2731894363/001/3667903834 --lineno 42 --lines-in-file 100 --time 1585598059 --hide-branch-names .* --project wakatime-cli --write --verbose: fork/exec ./build/wakatime-cli-linux-riscv64: no such file or directory
FAIL    command-line-arguments  0.085s
FAIL
==> ERROR: A failure occurred in check().
    Aborting...
```

wakatime/riscv64.patch
```diff
@@ -0,0 +1,34 @@
...
+-source=("${pkgname}-cli::git+https://github.com/${pkgname}/${pkgname}-cli#tag=v${pkgver}")
+-sha256sums=('SKIP')
++source=("${pkgname}-cli::git+https://github.com/${pkgname}/${pkgname}-cli#tag=v${pkgver}"
++        "main_test-arch.patch::https://github.com/wakatime/wakatime-cli/pull/645.patch")
++sha256sums=('SKIP'
++            'a53e0d001d2a8e0bcf1e7c0ceb82a82d0ece4b65663af56387d898fb5b2e1e83')
+ 
+-_binname="${pkgname}-cli-linux-amd64"
++_GOARCH=$(go env GOARCH)
++_binname="${pkgname}-cli-linux-${_GOARCH}"
```
piggynl, line 16: Hi, it appears that `go` has not been installed when this line is executed, which results in `_GOARCH` being an empty string. Do you have any suggestions for a solution?

cubercsl: Can we place these lines in `build()`?

piggynl: Thanks!

r-value: Actually I think it might be better to initialize the enviroment variable in `prepare()`?

r-value: Actually I think it might be better to initialize the enviroment variable in `prepare()`?

piggynl: Hi @r-value, by Arch Linux wiki page [Creating packages](https://wiki.archlinux.org/title/Creating_packages#prepare()), `prepare()` is used to prepare sources and will not be run if `--noextract` is specified for `makepkg`. However, in this case, the variable `_binname` should always be set regardless of whether the source code is extracted.

> **prepare()**
>
> With this function, commands that are used to prepare sources for building are run, such as [patching](https://wiki.archlinux.org/title/Patching_in_ABS). This function runs right after package extraction, before [pkgver()](https://wiki.archlinux.org/title/Creating_packages#pkgver()) and the build function. If extraction is skipped (makepkg --noextract), then prepare() is not run.

r-value: > Hi @r-value, by Arch Linux wiki page [Creating packages](https://wiki.archlinux.org/title/Creating_packages#prepare()), `prepare()` is used to prepare sources and will not be run if `--noextract` is specified for `makepkg`. However, in this case, the variable `_binname` should always be set regardless of whether the source code is extracted.
> 
> > **prepare()**
> > With this function, commands that are used to prepare sources for building are run, such as [patching](https://wiki.archlinux.org/title/Patching_in_ABS). This function runs right after package extraction, before [pkgver()](https://wiki.archlinux.org/title/Creating_packages#pkgver()) and the build function. If extraction is skipped (makepkg --noextract), then prepare() is not run.

True enough.

r-value (approved):

wakatime/riscv64.patch
```diff
@@ -0,0 +1,36 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -14,12 +14,21 @@ license=('BSD')
+ depends=('glibc')
+ makedepends=('git' 'go')
+ checkdepends=('svn')
+-source=("${pkgname}-cli::git+https://github.com/${pkgname}/${pkgname}-cli#tag=v${pkgver}")
+-sha256sums=('SKIP')
++source=("${pkgname}-cli::git+https://github.com/${pkgname}/${pkgname}-cli#tag=v${pkgver}"
++        "main_test-arch.patch::https://github.com/wakatime/wakatime-cli/pull/645.patch")
```
felixonmars, line 10: The patch name looks general enough that may conflict with others when `SRCDEST` is set. Please prefix it with something like `$pkgname-`.

felixonmars (changes requested):

PR #722, opened 2022-02-05
addpkg: python-pyicu

Ast-x64:


Xeonacid:
This has been fixed in Arch.

PR #723, opened 2022-02-05
addpkg: dsniff

Ast-x64:


Xeonacid:
This has been fixed in Arch.

PR #739, opened 2022-02-05
addpkg: qbittorrent

Ast-x64:


r-value:
Confirmed that the best solution is to patch `libtorrent`, which directly referenced to pthread, instead of patching `qbittorrent`. This patch can be removed once `libtorrent` releases the change in arvidn/libtorrent#6722 and rebuilt.

PR #747, opened 2022-02-06
addpkg: ccls

r-value:
Ref: https://bugs.llvm.org/show_bug.cgi?id=51247


CoelacanthusHex:
@r-value LLVM bugzilla migrate to GitHub，use GitHub link better.
https://github.com/llvm/llvm-project/issues/50591

PR #751, opened 2022-02-06
addpkg: python-clickclick

Ast-x64:


python-clickclick/riscv64.patch
```diff
@@ -0,0 +1,13 @@
+diff --git PKGBUILD PKGBUILD
+index dd4af1716..910afe88c 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -8,7 +8,7 @@ url='https://codeberg.org/hjacobs/python-clickclick'
+ license=('Apache')
+ arch=('any')
+ depends=('python-click' 'python-yaml')
+-makedepends=('python-setuptools' 'flake8')
++makedepends=('python-setuptools' 'flake8' 'python-six')
```
felixonmars, line 10: It seems the project mentions `six` only in setup.py. Probably it should be removed.

felixonmars: PR'ed as https://codeberg.org/hjacobs/python-clickclick/pulls/1

PR #762, opened 2022-02-07
addpkg: oxipng

XieJiSS:


XieJiSS:
FYI `CFLAGS+=" -Wl,-plugin-opt=-target-abi=lp64d"` doesn't work here.

PR #773, opened 2022-02-07
addpkg: gn

Avimitin:
ld can't link soft-float modules with double-float modules with lto enabled. This package can be built with the lto option off. But lto is not the main reason for the build failure. So this patch can be removed when lto is correctly handling the float module.


Avimitin:
Ref: https://github.com/llvm/llvm-project/issues/50591

PR #776, opened 2022-02-08
addpkg: lldb

XieJiSS:
Ref: https://reviews.llvm.org/D102872


lldb/riscv64.patch
```diff
@@ -0,0 +1,33 @@
...
++  # lldb-server does not support riscv64
++  patch -Np1 -i ../0000-disable-lldb-server.patch
++
+   mkdir build
+ }
+ 
+ build() {
+   cd "$srcdir/$pkgname-$pkgver.src/build"
+ 
++  CFLAGS="$CFLAGS -pthread -latomic"
```
felixonmars, line 29: Are both `-pthread` and `-latomic` really needed here?

XieJiSS: Fixed. `-latomic` is not necessary if `-pthread` presents.

PR #779, opened 2022-02-08
upgpkg: alertmanager

XieJiSS:
https://bugs.archlinux.org/task/73677


Xeonacid:
This package can be built successfully now without any patch, thus should be removed.

PR #792, opened 2022-02-09
fixpkg: sympol

r-value:
Fix uninstalled dependency: `bliss` and bump pkgrel to 10.1
`namcap` reports reference to `libbliss.so` from `libsympol.so.0.1.9`


felixonmars:
Reported as https://bugs.archlinux.org/task/74284

PR #794, opened 2022-02-10
addpkg: ibus-unikey

xctan:


felixonmars:
Do you think this change is specific to RISC-V? Please also report it to Arch if that is not the case :)

PR #798, opened 2022-02-10
rmpkg:  libretro-kronos

rapiz1:
Upstreamed in https://github.com/libretro/yabause/commit/1e33fcad8f9bc7520e61f650e79d6e48ce163178


rapiz1:
My fault. Further investigation shows that there are two packages build from the same repository, but from different branch.

rapiz1:
For the record: https://github.com/libretro/yabause/pull/288

PR #806, opened 2022-02-11
addpkg: rssguard

cubercsl:
Maybe https://bugreports.qt.io/browse/QTBUG-73834


cubercsl:
https://github.com/felixonmars/archriscv-packages/blob/6a52354e195552706ea49156bca13e415d0d64de/qt5-base/riscv64.patch#L8

Fixed in https://github.com/felixonmars/archriscv-packages/commit/6a52354e195552706ea49156bca13e415d0d64de

PR #811, opened 2022-02-11
addpkg: gnuradio

r-value:


r-value:
Additional notes:
If we build `gnuradio` with LTO, `ld` complains about an undefined reference to `gr::digital::map_bb_impl::s_map_size`, which is a static variable with type of `static constexpr size_t`, and only referenced in the same translation unit as defined. This error disappears with LTO disabled so I'm attributing this to LTO.

PR #834, opened 2022-02-13
addpkg: stylelint

XieJiSS:


XieJiSS:
I've confirmed that 30s is enough for both qemu and sifive unmatched boards.

PR #841, opened 2022-02-15
addpkg: ksysguard

cubercsl:
The validgpgkey has been changed since https://github.com/archlinux/svntogit-packages/commit/08e076c13c6dd6e4471a5f4672b5d7c87907c439

[FS#73800](https://bugs.archlinux.org/task/73800)


cubercsl:
Upstream has fixed in https://github.com/archlinux/svntogit-packages/commit/a89bc47d6469465d19bf415362c151802fca1e09

PR #867, opened 2022-02-19
addpkg: cross

Ast-x64:


cross/riscv64.patch
```diff
@@ -0,0 +1,36 @@
...
+   cd ${pkgname}-${pkgver}
+-  cargo build --release --locked
++  cargo build --release
+ }
+ 
+ check() {
+   cd ${pkgname}-${pkgver}
+-  cargo test --release --locked
++  cargo test --release
+ }
```
r-value, line 34: Removing `--locked` isn't a good idea. Use `cargo update -p` instead to minimize the change and keep reproducibility.

```suggestion
+prepare() {
+  cd ${pkgname}-${pkgver}
+  patch -Np1 -i "../use_new_nix.patch"
+  cargo update -p nix
+}
 
 build() {
   cd ${pkgname}-${pkgver}
```

r-value (changes requested):

SpriteOvO:
This package can now be built without a patch.

PR #871, opened 2022-02-19
addpkg: vinagre

Ast-x64:


r-value:
Any further explanation about this change? How `configure: error: spice support requested but not found` relates to dependency `gobject-introspection`?

FantasqueX:
It is clear after reading config.log.
```
configure:5386: checking for SPICE
configure:5393: $PKG_CONFIG --exists --print-errors "$SPICE_DEPS"
configure:5396: $? = 0
configure:5410: $PKG_CONFIG --exists --print-errors "$SPICE_DEPS"
configure:5413: $? = 0
configure:5427: result: no
Package 'gobject-introspection-1.0', required by 'spice-client-glib-2.0', not found
configure:5478: error: spice support requested but not found
```
I reported to ArchLinux as well. https://bugs.archlinux.org/task/76597

PR #874, opened 2022-02-19
addpkg: guitarix

Ast-x64:


r-value (approved):
Disabling LTO is necessary. Bug reported in [FS#74270](https://bugs.archlinux.org/task/74270).

PR #906, opened 2022-02-20
addpkg: yabause-gtk

xctan:


piggynl:
Note for reviewer: This patch is a workaround for `relocation truncated to fit: R_RISCV_JAL against XX` from gcc.

PR #931, opened 2022-02-22
addpkg: sbxkb

xctan:


r-value (approved):
Build also fails on x86-64. Reported in [FS#74269](https://bugs.archlinux.org/task/74269)

PR #1006, opened 2022-03-21
upgpkg: nodejs-lts-gallium

XieJiSS:
6 test cases will fail in QEMU, but can pass on SiFive Unmatched boards.

See #582


XieJiSS:
https://github.com/nodejs/node/pull/41566 has been merged into the nodejs `v16.14.x` branch, so we don't need to apply it again. This PR removes the patch from the source array.

PR #1024, opened 2022-03-27
addpkg: cmake

Ast-x64:


r-value:
There are indications that pthread isn't the smoking gun but `std::atomic`. We should take more consideration.

PR #1034, opened 2022-03-28
addpkg: gnucash-docs

FantasqueX:


XieJiSS:
Maybe we can work on this after https://bugs.archlinux.org/task/73666 is fixed :-)

PR #1035, opened 2022-03-28
updpkg: qcoro

r-value:
+ Adapt patch to latest release
 + Better approach for atomic operations
 + Changes upstreamed in danvratil/qcoro#52


felixonmars:
Great job!

PR #1040, opened 2022-03-29
addpkg: bacon

FantasqueX:
According to https://riscv-notes.sh1mar.in/docs/record/rust-related#riscv64gc-unknown-linux-gnu

### DIFF
--- a/bacon/riscv64.patch
+++ b/bacon/riscv64.patch
@@ -1,0 +1,13 @@
+diff --git a/trunk/PKGBUILD b/trunk/PKGBUILD
+index be2c48c..6b1be7b 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -24,7 +24,7 @@ pkgver() {
+ prepare() {
+   cd "$pkgname"
+ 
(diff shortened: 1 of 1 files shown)


Avimitin:
This can be simply done by removing the `--target` argument. Ref: https://github.com/felixonmars/archriscv-packages/pull/1025/commits/1a2691a1f5a42cbbe375cfaa15e6ae6f12b100ac

bacon/riscv64.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/trunk/PKGBUILD b/trunk/PKGBUILD
```
cubercsl, line 1:
```suggestion
diff --git PKGBUILD PKGBUILD
```

Avimitin, line 1: We should remove the prefix `a/trunk` and `b/trunk` here.

FantasqueX:
Fixed. Any other suggestions?

r-value (approved):
LGTM

PR #1041, opened 2022-03-29
fixpkg: fcitx5-mozc

sakuraikaede:
Fix wrong patch file


r-value:
Unfortunately, dependency `bazel` is unresolvable right now. This patch actually doesn't work.

```
==> Installing missing dependencies...
error: target not found: bazel
==> ERROR: 'pacman' failed to install missing dependencies.
==> Missing dependencies:
  -> python
  -> gtk2
  -> subversion
  -> bazel
  -> git
  -> clang
  -> python-six
==> ERROR: Could not resolve all dependencies.
```

XieJiSS:
Waiting for #1046 to be reviewed and merged.

XieJiSS:
This package has been built successfully on July 15, 2022, so I believe that this PR is not needed anymore :-) Feel free to reopen this if you think that this is a mistake!

PR #1044, opened 2022-03-30
upgpkg: amdvlk 2022.Q1.3-1

Ast-x64:
It is no longer required to use lld and clang because the linking error has been fixed. See https://github.com/GPUOpen-Drivers/llpc/issues/1645 and https://github.com/GPUOpen-Drivers/llpc/pull/1698.


XieJiSS:
`amdvlk` has been built successfully on Sep 12, 2022, so I suspect that this PR is not needed anymore :-)

PR #1048, opened 2022-03-30
addpkg: av1an

Avimitin:
This patch will fix two dependencies sources:

* Replace ffmpeg-next with my patched version

The `ffmpeg-next` crate relies on the `ffmpeg-sys-next` crate. And the
`ffmpeg-sys-next` misuses the arch type that LLVM can't recognize.

I've fixed this and made a PR for the `ffmpeg-sys-next`, but the
upstream is not active for nearly three months. So I have to replace
the crates.io source with my patched repository.

Reference: Avimitin/rust-ffmpeg-sys@3e2b72b

* Update serde to 1.0.132

The 1.0.132 version enables atomic64 serialization on RISC-V. And av1an
relies on this feature.

Reference: serde-rs/serde@0508cb5

Additional information: We must precisely update the `serde` and
`ffmpeg-next` crate. This may seem redundant but one of the dependencies
of the av1an, indicatif, didn't follow semver convention and make
breaking changes in the same major version. So we should not simply run
`cargo update` here.

### DIFF
--- a/av1an/fix-dependency.patch
+++ b/av1an/fix-dependency.patch
@@ -1,0 +1,41 @@
+diff --git Cargo.toml Cargo.toml
+index 4f2762c..d77cea4 100644
+--- Cargo.toml
++++ Cargo.toml
+@@ -18,7 +18,7 @@ name = "av1an"
+ [dependencies]
+ anyhow = "1.0.42"
+ serde_json = "1.0.64"
+-serde = { version = "1.0.126", features = ["serde_derive"] }
++serde = { version = "1.0.132", features = ["serde_derive"] }
+ shlex = "1.0.0"
+ path_abs = "0.5.1"
+ av1an-cli = { path = "av1an-cli", version = "0.3.1" }
+diff --git av1an-cli/Cargo.toml av1an-cli/Cargo.toml
+index afbc8eb..472cb8b 100644
+--- av1an-cli/Cargo.toml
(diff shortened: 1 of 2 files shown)


felixonmars:
> The `ffmpeg-next` crate relies on the `ffmpeg-sys-next` crate. And the
> `ffmpeg-sys-next` misuses the arch type that LLVM can't recognize.

In this case, can we [patch] `ffmpeg-sys-next` instead of replacing `ffmpeg-next` entirely?

Avimitin:
> > The `ffmpeg-next` crate relies on the `ffmpeg-sys-next` crate. And the
> > `ffmpeg-sys-next` misuses the arch type that LLVM can't recognize.
> 
> In this case, can we [patch] `ffmpeg-sys-next` instead of replacing `ffmpeg-next` entirely?

I've tried. The av1an contains two projects in the workspace. 

```text
.
|-av1an-core/
	|-Cargo.toml
|-av1an-cli/
	|-Cargo.toml
Cargo.toml
```

And cargo refused to patch the crate source for the sub projects. It prompts up error like: `patch for the non root package will be ignored, specify patch at the workspace root`. 

Also it doesn't take affect even if I patch the crate source in workspace root. It's like a paradox. I can't patch the sub package, and patching root has no effect. So fork those dependencies one by one is the only solution I can think of.

PR #1049, opened 2022-03-30
fixpkg: qbittorrent

sakuraikaede:


qbittorrent/riscv64.patch
```diff
@@ -21,3 +20,21 @@
  build() {
    cmake -B build -S $pkgbase-$pkgver \
      -DCMAKE_INSTALL_PREFIX=/usr \
+-    -DQT6=ON
+-  cmake --build build
++    -DQT6=ON \
++    -DCMAKE_EXE_LINKER_FLAGS="-latomic"
```
r-value, line 26: I suggest fixing this problem in upstream build system first.

sakuraikaede: OK. Changes upstreamed in qbittorrent/qBittorrent#16788

r-value:
Closing this because no patch is required for `qbittorrent` since `gcc-11.2.0-4.3` update.

PR #1056, opened 2022-04-01
addpkg: telegram-desktop

SpriteOvO:
Tried to fix it upstream ([telegramdesktop/tdesktop](https://github.com/telegramdesktop/tdesktop), PR 24275), but upstream refused to merge.

Explicitly link to `libatomic` via `CXXFLAGS` to fix the `undefined reference` error for atomic functions.

### DIFF
--- a/telegram-desktop/riscv64.patch
+++ b/telegram-desktop/riscv64.patch
@@ -1,0 +1,11 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -21,7 +21,7 @@ build() {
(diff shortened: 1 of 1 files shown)


XieJiSS:
Better to ref the upstream PR directly, so that possibly people reaching there can be guided to here https://github.com/telegramdesktop/tdesktop/pull/24275

XieJiSS:
Things have changed; since we are patching gcc's spec now, IMO this should be closed

PR #1058, opened 2022-04-02
addpkg: libvorbis

Ast-x64:
Force `-j1` to avoid racing on moving the `.Tpo` file.


r-value:
Is it possible to patch the `Makefile` to fix the file dependencies?

felixonmars:
Merging for now. This should still be at least reported upstream though :)

PR #1078, opened 2022-04-07
addpkg: python-dictpath

cubercsl:
`python-six` used to be installed as a dependency by `python-packaging`, but was removed in recent testing. https://github.com/archlinux/svntogit-packages/commit/a972071e011277914e65f46b2fe472a2de1ebb1a

Also FTBFS with `testing-x86_64-build` [FS#74382](https://bugs.archlinux.org/task/74382)


cubercsl:
Fixed in upstream, closed.

PR #1079, opened 2022-04-07
addpkg: symengine

piggynl:
This patch resolves the test timeouts on components that requires LLVM's MCJIT. (e.g. [tests/eval/test_lambda_double.cpp](https://github.com/symengine/symengine/blob/0cac84a08110b9047342ee12b443cd85f1698076/symengine/tests/eval/test_lambda_double.cpp), [tests/cwrapper/test_cwrapper.c](https://github.com/symengine/symengine/blob/0cac84a08110b9047342ee12b443cd85f1698076/symengine/tests/cwrapper/test_cwrapper.c))


felixonmars:
Is LLVM only used for the test? If that's not the case, disabling LLVM for only test failures doesn't look like a correct solution.

piggynl:
Hi @felixonmars, symengine only uses LLVM to compute floating point data via LLVM's MCJIT. However, MCJIT doesn't support RISC-V, so is it reasonable to disable LLVM in this case?

felixonmars:
> Hi @felixonmars, symengine only uses LLVM to compute floating point data via LLVM's MCJIT. However, MCJIT doesn't support RISC-V, so is it reasonable to disable LLVM in this case?

That's indeed a good reason. Thanks for investigating!

PR #1089, opened 2022-04-09
addpkg: libgweather-4

moodyhunter:
~~1. There have been problems with `pylint` while it's (probably?) worthless to check for linting when packaging.~~ (fixed in #1092)
2. Increased test timeout (qemu is slow, give it more time) (it's taking 41secs avg on my Laptop)


felixonmars:
Great! Please also try to upstream this change :)

PR #1090, opened 2022-04-09
fixpkg: zlib

Ast-x64:


hexchain:
Context:
- https://github.com/madler/zlib/issues/613
- https://bugs.archlinux.org/task/74371

PR #1092, opened 2022-04-09
addpkg: python-pylint

moodyhunter:
increase its test timeout, upstream change: https://github.com/PyCQA/pylint/pull/6241 (merged)


r-value:
Reproduced on 5950X builder.

felixonmars:
The path is actually wrong so the patch wasn't applied.

moodyhunter:
gonna fix it

PR #1134, opened 2022-04-15
addpkg: gdk-pixbuf2

Avimitin:
The "pixbuf-randomly-modified" test randomly modifies the file and then tries to load it, which needs vast amounts of RAM and then is killed by the SIGKILL signal. There are some similar issue reports on:

* Debian i386: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=942124
* FreeBSD-amd64: https://buildd.debian.org/status/fetch.php?pkg=gdk-pixbuf&arch=kfreebsd-amd64&ver=2.40.0%2Bdfsg-2&stamp=1582013991&raw=0
* Yocto: http://bugzilla.yoctoproject.org/show_bug.cgi?id=13840

The test error was reported on GNOME GitLab issues, and one of the contributors thinks they should replace the test. But two years have gone by, the issue is not active now. No contributor take action on it. So I decide to skip this test.

* Upstream Report: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/issues/146

### DIFF
--- a/gdk-pixbuf2/fix-sigkill.patch
+++ b/gdk-pixbuf2/fix-sigkill.patch
@@ -1,0 +1,13 @@
+diff --git tests/meson.build tests/meson.build
+index 7c6cb11..a6393f9 100644
+--- tests/meson.build
++++ tests/meson.build
+@@ -79,7 +79,7 @@ installed_tests = {
+   },
+   'pixbuf-fail': { 'suites': ['conform', 'slow'], },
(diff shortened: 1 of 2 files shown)


felixonmars:
I would suggest adding a comment on the upstream report, if this is reproduced reliably on real unmatched boards.

Avimitin:
Close this pr as this package can be built successfully on Unmatched board. This yet another qemu-user specific issue.

PR #1136, opened 2022-04-15
addpkg: qt6-webengine

moodyhunter:
could probably work, or at least it builds

mainly ported from the patch by openSUSE


moodyhunter:
I can't confirm its soundness, it would **probably** work, or, unfortunately, be broken.

Some test cases pass but some didn't, can't diagnose why.

PR #1141, opened 2022-04-16
addpkg: pyside2

Avimitin:
llvm-config crashed in qemu-user because it can't locate clang's built-in include directory (neither by checking the environment variables LLVM_INSTALL_DIR, CLANG_INSTALL_DIR  nor running llvm-config).

And it can be built without any modification in Unmatched board.

### DIFF
--- a/qemu-user-blacklist.txt
+++ b/qemu-user-blacklist.txt
@@ -40,6 +40,7 @@ perl
 perl-par-packer
 procs
 pueue
+pyside2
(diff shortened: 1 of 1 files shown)


Avimitin:
Close because it is done in commit 8e6d92f8d0df1f8ec535b4e902b7d7ea0f77a046

felixonmars:
Oops. Thanks for the details :)

PR #1143, opened 2022-04-16
addpkg: gdk-pixbuf2

Avimitin:
This patch add gdk-pixbuf2 into noqemu list and increase the timeout limit for "pixbuf-randomly-modified" test.

This package can't be built in qemu user environment. The "pixbuf-randomly-modified" test randomly modifies the file and then tries to load it. This test process needs vast amounts of RAM and then the test will be killed by the SIGKILL signal.

It can be built successfully on Unmatched board, but we still can't pass the "picbuf-randomly-modified" test. This time the failure is caused by timeout. 500s is a reasonable value for the Unmatched performance.

### DIFF
--- a/gdk-pixbuf2/fix-timeout.patch
+++ b/gdk-pixbuf2/fix-timeout.patch
@@ -1,0 +1,13 @@
+diff --git tests/meson.build tests/meson.build
+index 7c6cb11..7f88381 100644
+--- tests/meson.build
++++ tests/meson.build
+@@ -203,7 +203,7 @@ foreach test_name, test_data: installed_tests
+ 
+   # Two particularly slow tests
+   if test_suites.contains('slow')
+-    timeout = 300
++    timeout = 500
+   else
+     timeout = 30
+   endif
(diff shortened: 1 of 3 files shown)


Avimitin:
References:

- Upstream Report: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/issues/146
- Debian i386: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=942124
- FreeBSD-amd64: https://buildd.debian.org/status/fetch.php?pkg=gdk-pixbuf&arch=kfreebsd-amd64&ver=2.40.0%2Bdfsg-2&stamp=1582013991&raw=0
- Yocto: http://bugzilla.yoctoproject.org/show_bug.cgi?id=13840

felixonmars:
I would suggest to upstream the `500s` change since that's for a very valid reason.

Avimitin:
Track the timeout issue here: https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/issues/208

PR #1146, opened 2022-04-17
noqemu: add PyQt6 family

moodyhunter:
PyQt6 builder calls qmake, which in turn triggers the qemu bug.


moodyhunter:
This can be closed in favour of #1151

PR #1151, opened 2022-04-17
addpkg: qt6-base

moodyhunter:
Add compilation flag to disable `forkfd_pidfd`.

### DIFF
--- a/qt6-base/riscv64.patch
+++ b/qt6-base/riscv64.patch
@@ -1,0 +1,12 @@
+diff --git PKGBUILD PKGBUILD
+index 937ea28..b8c2d9e 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -55,6 +55,7 @@ build() {
+     -DQT_FEATURE_openssl_linked=ON \
+     -DQT_FEATURE_system_sqlite=ON \
+     -DQT_FEATURE_system_xcb_xinput=ON \
++    -DQT_FEATURE_forkfd_pidfd=OFF \
+     -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
+   cmake --build build
+ }


moodyhunter:
This would deprecate #1146

felixonmars:
We might want to add more explanation here, just in case :)

moodyhunter:
Qt, for its Unix QProcess implementation, utilises [forkfd library](https://doc.qt.io/qt-6/qtcore-attribution-forkfd.html), (See also: https://github.com/qt/qtbase/tree/dev/src/3rdparty/forkfd) as a helper library to fork subprocess. 

`pidfd` is a new kernel feature, which has just been added to the kernel upon v5.2.

Since [this merge](https://codereview.qt-project.org/c/qt/qtbase/+/313894), Qt added a new configuration feature (`forkfd_pidfd`, "CLONE_PIDFD support in forkfd") and it was default to ON on Linux.

When this feature is ON, Qt does nothing special, forkfd library selectively uses either `forkfd()` or `fork()` based on kernel version: (it goes into [this branch](https://github.com/qt/qtbase/blob/dev/src/3rdparty/forkfd/forkfd.c#L654-L656) and then [sets CLONE_PIDFD here](https://github.com/qt/qtbase/blob/dev/src/3rdparty/forkfd/forkfd_linux.c#L150) before calling [clone](https://github.com/qt/qtbase/blob/dev/src/3rdparty/forkfd/forkfd_linux.c#L68))

OTOH, if the feature is OFF, Qt sets the `FFD_USE_FORK` flag (say: forkfd please use `fork` nevertheless) when calling `::forkfd()`, which will cause forkfd to use `fork()` direcly.

Under regular circumstances, this works well, forkfd is smart enough to correctly detect the kernel version and use the correct syscall, however, things go south fast when `qemu-user` steps in:

1. Qt has no `FFD_USE_FORK` set, and
2. forkfd thinks the kernel has correct `forkfd` syscall implementation (because it should have), and
3. **QEMU has no support for `clone`-ing with `CLONE_PIDFD` yet (there's a [possibly abandoned patch](https://patchew.org/QEMU/mvm4kadwyrm.fsf@suse.de/))**

So, to save our boards from the `noqemu` hell, this PR adds the argument `QT_FEATURE_forkfd_pidfd=OFF` to CMake, allowing QProcess to work in our qemu-user environment, this would especially save `qmake`.

This PR also improves life expectancy by allowing us not to bother with the slow board build.

PR #1224, opened 2022-04-26
upgpkg: rust 1.60.0

Ast-x64:


rust/riscv64.patch
```diff
@@ -1,47 +1,81 @@
...
  source=(
-@@ -62,11 +62,9 @@ link-shared = true
+   "https://static.rust-lang.org/dist/rustc-$pkgver-src.tar.gz"{,.asc}
+   "https://github.com/llvm/llvm-project/releases/download/llvmorg-$_llvm_ver/compiler-rt-$_llvm_ver.src.tar.xz"{,.sig}
++  "git+https://github.com/rust-lang/libc.git#tag=0.2.116"
+   0001-cargo-Change-libexec-dir.patch
+   0001-bootstrap-Change-libexec-dir.patch
+   0002-compiler-Change-LLVM-targets.patch
+   0003-compiler-Use-wasm-ld-for-wasm-targets.patch
++  "use_local_libc_cc.patch"
```
felixonmars, line 31: Looks like...this patch wasn't committed, and the committed "replace_patched_libc-cc.patch" wasn't used.

felixonmars (changes requested):

PR #1227, opened 2022-04-26
upgpkg: fcitx-mozc

SpriteOvO:
This patch fixed error:

```
Hard-float 'd' ABI can't be used for a target that doesn't support the D instruction set extension (ignoring target-abi)
/usr/bin/ld: /tmp/lto-llvm-208020.o: can't link soft-float modules with double-float modules
/usr/bin/ld: failed to merge target specific data of file /tmp/lto-llvm-208020.o
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
```

It cannot be fixed by `-Wl,-plugin-opt=-target-abi=lp64d` or `-mabi=lp64d`, so disable LTO.


felixonmars:
It appears to be updating the wrong folder.

SpriteOvO:
@felixonmars Oh, my bad. Sorry.

PR #1230, opened 2022-04-27
addpkg: fcitx-mozc

SpriteOvO:
See #1227.

This PR reverts the changes made in the wrong directory and puts them in the correct directory.

## UPDATE

`clang` seems to ignore ` -Wl,` in `CFLAGS` and `CXXFLAGS`, but not for `LDFLAGS`. So we can use `LDFLAGS="${LDFLAGS} -Wl,-plugin-opt=-target-abi=lp64d"` to make it work without disabling LTO.


SpriteOvO:
@felixonmars Squashed 2 commits into a single commit.

felixonmars:
> @felixonmars Squashed 2 commits into a single commit.

Thanks. Unfortunately CI is still failing :(

felixonmars:
Skipping CI as it's flagging the restored fcitx5-mozc folder.

PR #1238, opened 2022-04-29
fixpkg: typescript-language-server

piggynl:
This change fixes timeout in pre- and post-test hook.


piggynl:
Use `.mocharc.json` to specify timeout because it has the lowest priority. Any future upstream change will override this and we will notice the change.

Issue #1258, opened 2022-05-05
Mesa useless on D1 due to "draw-use-llvm=false"


sehraf:
Since the addition of `-D draw-use-llvm=false` in ac8df15198cf859bcf381bd56ef8c033d3faeb4e i get like 10px/s (pixel being updated per second, i can literally watch how the image is composed pixel after pixel :eyes: ) on an Allwinner D1.
I'm aware that this was changed due to other issues, i just want to point out that the D1 is currently useless when it comes to X/wayland. When compiling mesa with `-D draw-use-llvm=true` X works as before.

felixonmars:
Please give a try to 22.0.2-1.2 and let me know if it makes any difference. I am still unsure of what the right workaround should be, either :(

sehraf:
Thanks for looking into this!
Sadly, mesa 22.0.2-1.2 isn't helping :(

Here is an image when a (white) frame is updated, it slowly fills a triangle top to bottom:
![image](https://user-images.githubusercontent.com/2230104/167249547-6df77e5d-21fb-44bf-ba3d-2755a2dcdb12.png)

felixonmars:
I see. This means that llvmpipe is indeed what makes the difference here, unfortunately :(

We'll be looking into how to properly address this.

felixonmars:
Update: There is a [mesa patch](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/17801) available which requires LLVM 15+. We'll try this once Arch x86 has update to LLVM 15 (upstream release already available).

neednewuser:
Is there a way I can help speed this up? I saw LLVM has been updated to version 15+.

felixonmars:
> Is there a way I can help speed this up? I saw LLVM has been updated to version 15+.

Looks like no one is actively working on the orcjit patch yet :(

kxxt:
We will get LLVM OrcJIT support in mesa soon, see #3207 

![image](https://github.com/felixonmars/archriscv-packages/assets/18085551/cc0a8347-0904-4b07-b01d-734fd4f95225)

PR #1262, opened 2022-05-07
mesa: try to use a different approach for disabling llvmpipe

felixonmars:
A hack to disable llvmpipe without disabling draw-use-llvm entirely.

This is an attempt to investigate #1258 as I'm not sure how D1 benefits from draw-use-llvm.


felixonmars:
Turns out it doesn't solve anything. Closing.

PR #1264, opened 2022-05-11
addpkg: xine-lib

Avimitin:
Dav1d 1.0.0 break compatibility:
https://github.com/videolan/dav1d/commit/753eef833bdd8ff1585c5c858cafeca8fefbb16e


Avimitin:
This patch is copied from https://sourceforge.net/p/xine/tickets/11/, however there is no version released yet since this patch is merged.

PR #1269, opened 2022-05-12
addpkg: python-pbr

cubercsl:
set default `OS_TEST_TIMEOUT=60` suggested in upstream `tox.ini`.

https://github.com/openstack/pbr/blob/5.9.0/tox.ini#L16


cubercsl:
If env `OS_TEST_TIMEOUT` is not set, it will default to `30` in testcase.

https://github.com/openstack/pbr/blob/5.9.0/pbr/tests/base.py#L74

PR #1275, opened 2022-05-16
addpkg: v2ray

cvttlay:


cvttlay:
This patch is also report to upstream as PR v2fly/v2ray-core#1786 and issue v2fly/v2ray-core#1830.

PR #1278, opened 2022-05-16
upgpkg: ostree

SpriteOvO:


r-value (approved):
Patch context changes only, LGTM.

PR #1279, opened 2022-05-16
updpkg: gcc

r-value:
Adapt to gcc 12.1.0-1

Patches added in #1050 are removed because of the new release.


CoelacanthusHex:
arch=(riscv64) 带进来了

PR #1297, opened 2022-05-22
addpkg: wolfssl

cubercsl:
EdDSA448 requires SHAKE256 which requires SHA-3.

SHA-3 is default enabled in x86_64/aarch64 but not enabled in riscv64 by the CMakeLists.

https://github.com/wolfSSL/wolfssl/blob/v5.3.0-stable/CMakeLists.txt#L506-L513

### DIFF
--- a/wolfssl/riscv64.patch
+++ b/wolfssl/riscv64.patch
@@ -1,0 +1,12 @@
+diff --git PKGBUILD PKGBUILD
+index 186a0aaa..a0a61694 100644
+--- PKGBUILD
++++ PKGBUILD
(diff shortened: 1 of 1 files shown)


felixonmars:
We probably should know why it's not enabled for riscv64 though :)

SpriteOvO:
> We probably should know why it's not enabled for riscv64 though :)

FYI, default conditions are here https://github.com/wolfSSL/wolfssl/blob/e722c15be860794179082a05d09e6a90dc77ccf0/configure.ac#L2390-L2398

PR #1303, opened 2022-05-23
updpkg: pacman 6.0.1-5

r-value:
No changes in `makepkg.conf`.

### DIFF
--- a/pacman/riscv64.patch
+++ b/pacman/riscv64.patch
@@ -1,5 +1,5 @@
 diff --git PKGBUILD PKGBUILD
-index 5321ff1b..b20f85ee 100644
+index 8ca61cf4..1317f94c 100644
 --- PKGBUILD
 +++ PKGBUILD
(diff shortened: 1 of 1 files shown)


pacman/riscv64.patch
```diff
@@ -25,14 +26,6 @@ index 5321ff1b..b20f85ee 100644
  
  
  prepare() {
-@@ -43,6 +45,7 @@ prepare() {
-   patch -Np1 -i ../makepkg-use-ffile-prefix-map-instead-of-fdebug-prefi.patch
-   patch -Np1 -i ../libmakepkg-add-extra-buildflags-only-when-buildflags.patch
-   patch -Np1 -i ../make-link-time-optimization-flags-configurable.patch
-+  patch -Np0 -i ../makepkg-config.guess.patch
```
felixonmars, line 32: If we are going to remove this hack altogether, we should also remove the patch itself and from the source array.

felixonmars (changes requested):

PR #1313, opened 2022-05-25
addpkg: gss

cubercsl:


felixonmars:
We could also copy our own packaged version of these files from the system:

/usr/share/autoconf/build-aux/config.guess
/usr/share/automake-1.16/config.guess
/usr/share/libtool/build-aux/config.guess

CoelacanthusHex:
As upstream manual said, we can use one line command to update these files:
```
automake --add-missing --copy --force-missing
```

CoelacanthusHex:
There is also another command：
```
autoreconf -fi
```
It is used by many upstream and distribution such as debian, AOSC OS

PR #1316, opened 2022-05-26
addpkg: python-doit-py-bootstrap

IEAST:
1. `pyton-doit-py` depends on `python-doit`, and `python-doit`  checkdepends on `python-doit-py`
2. build order: ` python-doit-py` -> `python-doit` -> `python-doit-py`


felixonmars:
> 1. `pyton-doit-py` depends on `python-doit`, and `python-doit`  checkdepends on `python-doit-py`

Doing nocheck builds will solve this more easily, in this order:

`python-doit:nocheck -> python-doit-py -> python-doit`

PR #1323, opened 2022-05-27
addpkg: libxv

FantasqueX:
This package failed to build because config.guess and config.sub bundled
in release tarball are too old. So force autoreconf in prepare() should
solve this problem. A makedepends is added due to requirements of
autoreconf.

Also, a bug has been filed to upstream.
https://bugs.archlinux.org/task/74866


FantasqueX:
FYI, this issue has been reported to Xorg upstream. https://gitlab.freedesktop.org/xorg/lib/libxv/-/issues/1

PR #1325, opened 2022-05-27
addpkg: libxxf86vm

FantasqueX:
This package failed to build because config.guess and config.sub bundled
in release tarball are too old. So force autoreconf in prepare() should
solve this problem.

Also, a bug has been filed to upstream.
https://bugs.archlinux.org/task/74869


Xeonacid:
This should be reported to the real upstream instead of Arch.

FantasqueX:
Hi! I agree with reporting this problem to Xorg. Here is the upstream issue.

https://gitlab.freedesktop.org/xorg/lib/libxxf86vm/-/issues/1

However, I don't think this problem will be fixed due to inactive upstream. So this patch is an effective one currently.

PR #1330, opened 2022-05-28
upgpkg: rustypaste

SpriteOvO:
The performance of QEMU RISC-V is lower than the assumptions in tests `test_upload_expiring_file` and `test_get_expired_files`. This patch increases the expiration time in these tests.


rustypaste/riscv64.patch
```diff
@@ -1,22 +1,23 @@
...
    cd ${pkgname}-${pkgver}
 -  cargo fetch --locked --target "$CARCH-unknown-linux-gnu"
-+  cargo fetch --locked
- }
- 
- build() {
-@@ -24,6 +24,8 @@ build() {
-   export RUSTUP_TOOLCHAIN=stable
-   export CARGO_TARGET_DIR=target
-   CFLAGS+=' -ffat-lto-objects'
```
XieJiSS, line 17: 这个 flag 确实不需要了吗

XieJiSS: 还有上面两个 `export` 也是？

SpriteOvO: @XieJiSS 这些是 unchanged 部分，还在原文件中。我把 ring patch 从 `build()` 挪到 `prepare()` 了，所以 diff 有些区别。

felixonmars:
> The performance of QEMU RISC-V is lower

Do these tests pass on real hardware? I think we should notify upstream if so, or otherwise add this package to qemu-user-blacklist.

PR #1352, opened 2022-06-01
updpkg: qt5-webengine 5.15.9

r-value:
Also removes unused patch.


felixonmars:
IIRC that patch was there to record what was changed from the original v8 patch?

r-value:
Oops, committed it back. >_<

PR #1353, opened 2022-06-01
addpkg: elfutils

FantasqueX:
I got only three test suite failures when building on boards which are much
less than building by qemu-user. Those three failures are known to upstream and
this issue has been confirmed and assigned. Since it's tedious to investigate
why each tests failed, I just simply add elfutils to qemu-user-blacklists.txt.

Ref: https://sourceware.org/bugzilla/show_bug.cgi?id=29073

### DIFF
--- a/qemu-user-blacklist.txt
+++ b/qemu-user-blacklist.txt
@@ -10,6 +10,7 @@ dbus
 dovecot
 dqlite
 element
+elfutils
 ell
 ethtool
 fd


hexchain:
Related glibc change: https://sourceware.org/git/?p=glibc.git;a=commit;h=0b6c6750732483b4d59c2fcb45484079cd84157d

FantasqueX:
Superseded by #1590

PR #1362, opened 2022-06-02
addpkg: perl-file-rsyncp

TinySnow:
Fixed config.guess.

This package contains outdated config.guess file and couldn't correctly
guess system type. This patch is a temporary workaround that copy the
local config.guess file into project.

This is the same problem as some previous packages.
Ref: https://github.com/felixonmars/archriscv-packages/pull/1313.
Build passed.

### DIFF
--- a/perl-file-rsyncp/riscv64.patch
+++ b/perl-file-rsyncp/riscv64.patch
@@ -1,0 +1,12 @@
+diff --git PKGBUILD PKGBUILD
+index a5c6bcc..dcbc0ef 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -17,6 +17,7 @@ sha512sums=('b81d9ce63181605939e8aff87c9e56ab276899f019f44eb3f149e7ee0d44a94fcbe
+ 
+ prepare() {
+   cd File-RsyncP-$pkgver
++  cp -vf /usr/share/autoconf/build-aux/config.guess FileList/
+   # https://rt.cpan.org/Public/Bug/Display.html?id=141822
+   patch -Np1 -i ../gcc12.patch
+ }


perl-file-rsyncp/riscv64.patch
```diff
@@ -0,0 +1,12 @@
+diff --git PKGBUILD PKGBUILD
+index a5c6bcc..5292eca 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -24,6 +24,7 @@ prepare() {
+ build() {
+   cd File-RsyncP-$pkgver
+   USE_DEFAULT=1 perl Makefile.PL INSTALLDIRS=vendor
++  cp -vf /usr/share/autoconf/build-aux/config.guess FileList/
```
felixonmars, line 9: If the file wasn't generated by the above Makefile.PL call (it's unlikely, but if so, please fix the file used to generate it), this should be moved to prepare().

felixonmars (changes requested):

perl-file-rsyncp/riscv64.patch
TinySnow: Fixed it.
I'll notice it in the future, thanks.

perl-file-rsyncp/riscv64.patch
```diff
@@ -0,0 +1,12 @@
+diff --git PKGBUILD PKGBUILD
+index a5c6bcc..0a0efd6 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -18,6 +18,7 @@ sha512sums=('b81d9ce63181605939e8aff87c9e56ab276899f019f44eb3f149e7ee0d44a94fcbe
+ prepare() {
+   cd File-RsyncP-$pkgver
+   # https://rt.cpan.org/Public/Bug/Display.html?id=141822
```
felixonmars, line 8: This comment is about the gcc12.patch below, please don't insert an irrelevant fix in between :)

felixonmars (changes requested):

PR #1386, opened 2022-06-11
upgpkg: systemd

moodyhunter:
also add systemd into qemu-user-blacklist?


moodyhunter:
In my opinion, as to which one should this kind of change be called, either "upgpkg" or "fixpkg", should be discussed and altered in order to prevent future confusion.

PR #1389, opened 2022-06-12
addpkg: coreutils

FantasqueX:
This package fails to build on a riscv64 real hardware due to the same
reason on x86_64. The test-getlogin doesn't handle the case
/proc/self/loginuid is -1. After being reported to the upstream gnulib,
a patch to resolve this issue has been merged on master.
Ref: https://lists.gnu.org/archive/html/bug-gnulib/2022-06/msg00001.html

I also filed a bug on ArchLinux bug tracker which backports the patch to
resolve the issue of FTBFS.
Ref: https://bugs.archlinux.org/task/75112

Currently, I'm not sure why some unit tests fail when running with
qemu-user. Failing tests are listed below.
- FAIL: tests/misc/tac-2-nonseekable.sh
- FAIL: tests/tail-2/follow-stdin.sh

### DIFF
--- a/coreutils/fix-test-getlogin.patch
+++ b/coreutils/fix-test-getlogin.patch
@@ -1,0 +1,68 @@
+--- a/gnulib-tests/test-getlogin.h
++++ b/gnulib-tests/test-getlogin.h
+@@ -18,6 +18,7 @@
+ 
+ #include <errno.h>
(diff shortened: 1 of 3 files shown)


FantasqueX:
This PR is ready for review now🎉

PR #1396, opened 2022-06-13
addpkg: ninjas2

moodyhunter:
upstream has already updated the submodule, but idk when a new release will come :(


moodyhunter:
for more information: 
Its `BASE_OPTS` contains an unsupported `-mtune=` argument. 
- The patch removes such an erroneous argument for riscv64 while keeping it there for other architectures

PR #1397, opened 2022-06-14
addpkg: libepoxy

piggynl:
This patch fixes a wrapper for `dlsym()` in test, which hard-coded the glibc ABI version that libdl uses.

Upstream PR: anholt/libepoxy#270

### DIFF
--- a/libepoxy/riscv64.patch
+++ b/libepoxy/riscv64.patch
@@ -1,0 +1,23 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -14,8 +14,10 @@ makedepends=(git meson mesa-libgl systemd doxygen graphviz)
+ checkdepends=(xorg-server-xvfb)
+ options=(debug)
+ _commit=c84bc9459357a40e46e2fec0408d04fbdde2c973  # tags/1.5.10^0
+-source=("git+https://github.com/anholt/libepoxy#commit=$_commit")
+-sha256sums=('SKIP')
++source=("git+https://github.com/anholt/libepoxy#commit=$_commit"
++        "$pkgname-dlwrap-riscv64.patch::https://github.com/anholt/libepoxy/pull/270.patch")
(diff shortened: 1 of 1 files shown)


libepoxy/riscv64.patch
```diff
@@ -0,0 +1,23 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -14,8 +14,10 @@ makedepends=(git meson mesa-libgl systemd doxygen graphviz)
+ checkdepends=(xorg-server-xvfb)
+ options=(debug)
+ _commit=c84bc9459357a40e46e2fec0408d04fbdde2c973  # tags/1.5.10^0
+-source=("git+https://github.com/anholt/libepoxy#commit=$_commit")
+-sha256sums=('SKIP')
++source=("git+https://github.com/anholt/libepoxy#commit=$_commit"
++        "$pkgname-dlwrap-riscv64.patch::https://github.com/anholt/libepoxy/pull/270.patch")
```
XieJiSS, line 10: Just wondering, are similar patches necessary for other glibc versions, *e.g.* 2.26?

piggynl: No. libdl has different glibc ABI version for each architecture (e.g. rv64 has the version `GLIBC_2.27`, see [sysdeps/unix/sysv/linux/riscv/rv64/libdl.abilist](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/riscv/rv64/libdl.abilist;h=33ff573df8748197102314b6ca7579e060c7b8ad) in glibc's source), and it's not likely to update frequently. libepoxy uses a wrapper for `dlsym()`, which search through the hard-coded versions ([link](https://github.com/anholt/libepoxy/blob/a96abf1a4d29f78034273cbce96006cde950390c/test/dlwrap.c#L219-L274)) to resolve `dlsym()` provided by libdl.

PR #1438, opened 2022-06-27
qemu-user-blacklist: add glib-perl

moodyhunter:
`t/9.t:233:13` called fork() 🍴 while qemu-user hates forking.
`fork()` was reported successful but actually, it's not.


moodyhunter:
why it's a draft PR

PR #1440, opened 2022-06-28
upgpkg: android-tools

SpriteOvO:
Fixed rotten patch.


SpriteOvO:
Upstream patch in PR https://github.com/nmeum/android-tools/pull/66.

Arch Linux upstream's `PKGBUILD` download the `.tar.xz` file as source, which doesn't contain `patches` directory and all patch sets have been applied. If we pick that PR in `PKGBUILD`, then we have to replace the source with `.git` URL, and update submodules manually (this will take a long time). This is a little dirty, let's just keep the old patch in this repo and wait for upstream to release a new version.

PR #1441, opened 2022-06-28
addpatch: curl

moodyhunter:
the upstream release is missing an include, fixed in https://github.com/curl/curl/commit/e2e7f54b7bea521fa8373095d0f43261a720cda0.patch


moodyhunter:
now ready for review

PR #1446, opened 2022-06-29
qemu-user-blacklist: add pyqt6 family

moodyhunter:
pyqt6 needs qmake, this is broken in qemu-user


felixonmars:
Actually this file should list pkgbases instead of pkgnames. Sorry this wasn't properly documented.

moodyhunter:
is it just to remove the `python-` prefix?

felixonmars:
Probably true for the whole pyqt6 family.

PR #1461, opened 2022-07-04
Add gdc 11 for ldc bootstrapping

Ast-x64:
Files based on gcc and gcc11.

Disabled bootstrapping and checking progress for convenience.
`gdmd` is included together.


Xeonacid:
`gcc-d` may be a better pkgname as the naming convention. Adding patch to `gcc11` instead of a new pkgbase can also be considered.

PR #1469, opened 2022-07-08
addpatch: gnupg

SpriteOvO:
Update autotool configurations.

This problem will be reported to upstream and updated here later, because my account is still waiting for approval.


SpriteOvO:
Reported to upstream: https://dev.gnupg.org/T6078

PR #1481, opened 2022-07-12
addpkg: stfl

sukanka:
original source is not available, use github mirror instead.

### DIFF
--- a/stfl/riscv64.patch
+++ b/stfl/riscv64.patch
@@ -1,0 +1,19 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -11,12 +11,14 @@
+ depends=('ncurses')
+ makedepends=('patch' 'swig')
+ changelog=$pkgname.changelog
+-source=(http://clifford.at/stfl/$pkgname-$pkgver.tar.gz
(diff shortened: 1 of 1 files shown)


SpriteOvO:
@sukanka Hi! Thank you for opening this PR.

It seems that the problem is not only specific to the RISC-V architecture, also to x86_64. And https://clifford.at/stfl is still accessible, but the file ([link](https://www.clifford.at/stfl/stfl-0.24.tar.gz)) is no longer available for download.

So, I think this repo is not the best place to submit this patch. It is recommended that you:

1. Try contacting the developer of this package, the owner of https://www.clifford.at/stfl, and ask them to fix the problem.
2. If you are unable to contact them, or they do not respond, report this problem and submit this patch to [ArchLinux Upstream](https://bugs.archlinux.org/?project=5&string=stfl).

cubercsl:
It seems that source link is broken, and upstream mailist link is also broken  (No DNS record) :(

XieJiSS:
Hi! It seems like this package has been built successfully on Dec 30, 2022, so I believe that this PR can be closed :-) Feel free to reopen it if you think that this is a mistake!

PR #1485, opened 2022-07-13
addpatch: m2r

Ast-x64:


Xeonacid:
The package was dropped in Arch.

PR #1486, opened 2022-07-13
addpatch: efibootmgr

Ast-x64:


felixonmars:
A (somewhat) similar change was rejected before at https://github.com/rhboot/efibootmgr/pull/129

I don't think we should ship this change without mentioning this again to upstream.

PR #1491, opened 2022-07-14
rmvpatch: ldc-bootstrap

ZenithalHourlyRate:
This package has been removed in Arch.


XieJiSS:
Note that this is triggered by an automated task which is still under development, so there might be false positives

PR #1498, opened 2022-07-14
addpkg: simutrans-pak128

iningz:
Fix the illegally formated URL (also FTBFS in x86_64).
Bug reported: https://bugs.archlinux.org/task/75317


XieJiSS:
Seems like the patch has been modified by hand, *e.g.* the arch= line is missing in the context, which probably lead to CI failure.

PR #1499, opened 2022-07-14
addpkg: ttf-ibm-plex

yining765:
Update the package sums (also FTBFS in x86_64)
Bug reported: https://bugs.archlinux.org/task/75318


Xeonacid:
CI says the patch failed to be applied. You are missing the `arch=('any')` line.
Are you creating or modifying the patch manually? If so, please be cautious:-P

PR #1500, opened 2022-07-14
addpkg: simutrans-pak128

yining765:
Fix the illegally formated source URL (also FTBFS in x86_64)
Bug reported: https://bugs.archlinux.org/task/75317


Xeonacid:
Same as https://github.com/felixonmars/archriscv-packages/pull/1499#issuecomment-1184486401.

PR #1527, opened 2022-07-18
updpatch: gn

xctan:


gn/riscv64.patch
```diff
@@ -1,12 +1,22 @@
...
+ sha256sums=('SKIP')
++options=(!lto)
+ 
+ pkgver() {
+   cd $pkgname
+@@ -20,6 +21,9 @@ pkgver() {
  
  build() {
    cd $pkgname
-+  export LDFLAGS="$LDFLAGS -Wl,-plugin-opt=-target-abi=lp64d"
```
felixonmars, line 9: The old hack no longer works? That's unfortunate :(

PR #1532, opened 2022-07-20
addpatch: carla

moodyhunter:
EEL doesn't have a native RISC-V implementation, so add `EEL_TARGET_PORTABLE` to use the software-based one


felixonmars:
Please also submit the patch to upstream :)

moodyhunter:
will do

PR #1561, opened 2022-08-05
fixpatch: glibc

moodyhunter:
fix glibc rotten patch


moodyhunter:
note: currently glibc needs `nocheck` for packaging

PR #1569, opened 2022-08-08
addpatch: python-levenshtein

moodyhunter:
should this be reported upstream?


moodyhunter:
for x86_64, the `lib.linux-x86_64-*` is correct, however, the corresponding directory on riscv64 is named `lib/`

moodyhunter:
closing this for now, `python-setuptools 1:61.3.1-1` fixes the problem.

PR #1597, opened 2022-08-14
addpkg: prometheus-wireguard-exporter

Trumeet:
The PKGBUILD patch fixes two issues:

 1. rustc(1) recognizes `riscv64gc' instead of `riscv64'. It fixes the issue by substituting `riscv64' with `riscv64gc' in the `--target' argument if $CARCH is `riscv64'.

 2. This package depends on `ring', which is not yet riscv64 ready. This patch works around this issue by forcing Cargo to use felixonmars/ring.

The patched version passes compiling and testing.

References:

 1. https://github.com/felixonmars/archriscv-packages/commit/1e0a422d3e1a1b879b31988aada28441c5f25223

 2. https://wiki.archlinux.org/title/Rust_package_guidelines#Prepare

 3. https://github.com/felixonmars/archriscv-packages/wiki/%E6%88%91%E4%BB%AC%E7%9A%84%E5%B7%A5%E4%BD%9C%E4%B9%A0%E6%83%AF

### DIFF
--- a/prometheus-wireguard-exporter/riscv64.patch
+++ b/prometheus-wireguard-exporter/riscv64.patch
@@ -1,0 +1,23 @@
+diff --git PKGBUILD PKGBUILD
+index a573d5d..e11db13 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -23,7 +23,17 @@ sha512sums=('9ff42deaf28a45be26cf0211b11878be873467bee0bdb128efec3c1a9d4f7214f53
+ 
+ prepare() {
+   cd "prometheus_wireguard_exporter-${pkgver}"
+-  cargo fetch --locked --target "$CARCH-unknown-linux-gnu"
++
++  # RISC-V Patches:
++
++  # The upstream `ring' package does not support riscv yet.
++  # This workaround comes from https://github.com/felixonmars/archriscv-packages/wiki/%E6%88%91%E4%BB%AC%E7%9A%84%E5%B7%A5%E4%BD%9C%E4%B9%A0%E6%83%AF
++  echo -e "\n[patch.crates-io]\nring = { git = 'https://github.com/felixonmars/ring', branch = '0.16.20' }" >> Cargo.toml
++  cargo update -p ring
++
++  # rustc(1) recognizes `riscv64gc' instead of `riscv64'. Thus,
++  # as the convention, we removed `--target'.
++  cargo fetch --locked
+ }
+ 
+ build() {


prometheus-wireguard-exporter/riscv64.patch
```diff
@@ -0,0 +1,32 @@
...
++    _target="${CARCH}gc-unknown-linux-gnu"
++
++    # The upstream `ring' package does not support riscv yet.
++    # This workaround comes from https://github.com/felixonmars/
++    # archriscv-packages/wiki/%E6%88%91%E4%BB%AC%E7%9A%84%E5%B7
++    # %A5%E4%BD%9C%E4%B9%A0%E6%83%AF
++    echo -e "\n[patch.crates-io]\nring = { git = 'https://github.com/felixonmars/ring', branch = '0.16.20' }" >> Cargo.toml
++    cargo update -p ring
++  fi
++  cargo fetch --locked --target $_target
```
Xeonacid, line 29:
```suggestion
+  cargo fetch --locked
```
We can just do like this as a convention. 😄

Trumeet: Thanks!

prometheus-wireguard-exporter/riscv64.patch
```diff
@@ -0,0 +1,32 @@
...
++
++  # RISC-V Patches:
++  # rustc(1) recognizes `riscv64gc' instead of `riscv64'. Thus,
++  # we need to manually substitute riscv64 with riscv64gc in 
++  # --target. The `--target' argument, according to Arch Wiki,
++  # "tells cargo to only fetch dependencies needed for the
++  # specific target platform being built, thus reducing downloads."
++  # Ref: https://wiki.archlinux.org/title/Rust_package_guidelines#Prepare
++  local _target="$CARCH-unknown-linux-gnu"
++  if test $CARCH = "riscv64"; then
```
XieJiSS, line 19: As Xeonacid has suggested, we can simply remove this if branch and do necessary things unconditionally.

XieJiSS (changes requested):

prometheus-wireguard-exporter/riscv64.patch
Trumeet: Thank you. I will fix it right now.

Trumeet: How about the `ring` patch? May I do it unconditionally as well?

XieJiSS: I think so, at least we are doing so previously

XieJiSS: For this part we are looking forward to make the ring upstream package fully functional, or find an equivalent alternative, so that all patches specifically for ring can be removed, which would greatly reduce the amount of patched packages.

IMO maybe when some days we finally give up on that we will come back to the conditionally patch path.

Trumeet: OK, thank you!

XieJiSS (approved):

prometheus-wireguard-exporter/riscv64.patch
```diff
@@ -0,0 +1,25 @@
...
+ prepare() {
+   cd "prometheus_wireguard_exporter-${pkgver}"
+-  cargo fetch --locked --target "$CARCH-unknown-linux-gnu"
++
++  # RISC-V Patches:
++
++  # The upstream `ring' package does not support riscv yet.
++  # This workaround comes from https://github.com/felixonmars/
++  # archriscv-packages/wiki/%E6%88%91%E4%BB%AC%E7%9A%84%E5%B7
++  # %A5%E4%BD%9C%E4%B9%A0%E6%83%AF
```
r-value, line 16: I think we shouldn't break the line in the middle of a link. This will break the link jump function of various terminals.

```suggestion
+  # This workaround comes from: 
+  # https://github.com/felixonmars/archriscv-packages/wiki/%E6%88%91%E4%BB%AC%E7%9A%84%E5%B7%A5%E4%BD%9C%E4%B9%A0%E6%83%AF
```

XieJiSS, line 16:
```suggestion
+  # This workaround comes from https://github.com/felixonmars/archriscv-packages/wiki/%E6%88%91%E4%BB%AC%E7%9A%84%E5%B7%A5%E4%BD%9C%E4%B9%A0%E6%83%AF
```

r-value (changes requested):

prometheus-wireguard-exporter/riscv64.patch
Trumeet: Thank you. It is now fixed.

Trumeet: Fixed.

PR #1601, opened 2022-08-14
addpatch: cifs-utils

moodyhunter:
invoke autoreconf with the `-f` option to update `config.guess` and `config.sub`


felixonmars:
The patch file should be named `riscv64.patch` instead of `cifs-utils.patch`

moodyhunter:
my mistake

PR #1646, opened 2022-08-29
addpatch: gap

Avimitin:
There are three vendor libraries using old config.guess file. This patch
add a batch update script for those config.guess file before building
the gap project.

### DIFF
--- a/gap/riscv64.patch
+++ b/gap/riscv64.patch
@@ -1,0 +1,18 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -51,6 +51,15 @@ prepare() {
+   cd CddInterface-*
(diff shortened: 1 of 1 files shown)


gap/riscv64.patch
```diff
@@ -0,0 +1,18 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -51,6 +51,15 @@ prepare() {
+   cd CddInterface-*
+   sed -e 's|/usr/include/cdd|/usr/include/cddlib|' -i configure.ac # Fix build with cddlib 0.94k
+   rm configure
++  cd ..
++
++# Update the config.guess file
```
XieJiSS, line 9: Consider adding more indentation to this line, so that the # character can be vertically aligned with its sibling lines.

XieJiSS (changes requested):

Avimitin:
It was built successfully in the latest version, no more patches are needed.

PR #1647, opened 2022-08-29
addpkg: python-doublex-expects

yining765:
Fix python ModuleNotFoundError.
Bug reported: https://bugs.archlinux.org/task/75729


felixonmars:
From the backtrace, `hamcrest` is the module where the `six` dependency should go, not this package.

felixonmars:
Closing as `python-doublex-expects` now builds fine without changes with `python-pyhamcrest 1.10.1-4`.

PR #1651, opened 2022-08-30
addpatch: borg

Avimitin:
The packager script won't generate "-cpython-" suffix for the build
directory in riscv64. The build artifacts is still the same.

### DIFF
--- a/borg/riscv64.patch
+++ b/borg/riscv64.patch
@@ -1,0 +1,11 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -30,7 +30,7 @@ build() {
+ }
+ 
+ check() {
+-  cd "$_pkgname-$pkgver/build/lib.linux-$CARCH-cpython-"*/
++  cd "$_pkgname-$pkgver/build/lib.linux-$CARCH"*/
+   PYTHONPATH=$PWD PYTHONDONTWRITEBYTECODE=1 pytest -k 'not benchmark'
+ }
+ 


Avimitin:
The reason of this issue is python-setuptools have different implementation of platform detection between 61.3.1(Latest version in Arch RV) and 62.3.4(Latest version in Arch x86_64).

First of all, the name of the build directory is defined by a build option: `build-lib`. If user doesn't modify it manually, this option is default assigned to value of the base build directory(`build/`) plus a platform specfier.
The platform specifier is consist with current OS name and a tag. And the value of the tag is the key point of the whole issue:

* In 61.3.1, python-setuptools use `sys.version_info[:2]` as tag name, which should be `3` and `10` in our riscv machine.

```python
>>> import sys
>>> sys.version_info
sys.version_info(major=3, minor=10, micro=7, releaselevel='final', serial=0)
>>> sys.version_info[:2]
(3, 10)
```

* However in 62.3.4, python-setuptools use `sys.implementation.cache_tag` as tag name, so we will get the below output in x86_64

```python
>>> import sys
>>> sys.implementation.cache_tag
'cpython-310'
```

So that's why there is no "cpython" suffix when we build this package on Arch RV.

Reference:
* 61.3.1(Arch RV Latest) implementation of the build directory specifier: https://github.com/pypa/setuptools/blob/v61.3.1/setuptools/_distutils/command/build.py#L84
* 62.3.4(Arch x86_64 Latest) implementation of the build directory specifier: https://github.com/pypa/setuptools/blob/v62.3.4/setuptools/_distutils/command/build.py#L85

PR #1653, opened 2022-08-31
addpatch: srs-state-thread

Avimitin:
This patch adds RISC-V CPU support from upstream:
https://github.com/ossrs/state-threads/pull/28.

The README.md file was changed after the latest version and the
original patch can't be applied on it, so a modified local patch is also
submit here.

### DIFF
--- a/srs-state-threads/riscv64.patch
+++ b/srs-state-threads/riscv64.patch
@@ -1,0 +1,23 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -10,13 +10,18 @@ arch=('x86_64')
+ url="https://github.com/ossrs/state-threads/"
+ license=('GPL2' 'MPL')
+ makedepends=('git')
+-source=("git+https://github.com/ossrs/state-threads/#commit=${_pkgcommit}")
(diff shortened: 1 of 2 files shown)


srs-state-threads/riscv64.patch
```diff
@@ -0,0 +1,29 @@
...
+   sed -ie "s|LDFLAGS     =|LDFLAGS     = ${ldFlags}|" Makefile
++
++  # add riscv support
++  patch -Ni "$srcdir/riscv-cpu-support.patch"
+ }
+ 
+ build() {
+   cd "${srcdir}"/state-threads
+-  make STATIC_ONLY=no linux-optimized
++  make STATIC_ONLY=no EXTRA_CFLAGS="-D__riscv" linux-optimized
```
felixonmars, line 26: `__riscv` should have been defined by the compiler already. Do we really need to add it here?

Avimitin: I will try if this is necessary.

Avimitin: Updated.

PR #1668, opened 2022-09-07
upgpkg: caps

hilaolu:
fix sha512sum


felixonmars:
Assuming this problem is also present on Arch, please open a bug report to Arch too :)

PR #1684, opened 2022-09-10
addpatch: hplip

FantasqueX:
This patch is taken from https://salsa.debian.org/printing-team/hplip/-/blob/debian/main/debian/patches/0028-Remove-ImageProcessor-binary-installs.patch

Since not linking this library won't cause any link error, I think it's safe to do so.


FantasqueX:
### Ackownledge

Thanks @r-value and @CoelacanthusHex working together on this package.

PR #1686, opened 2022-09-11
updpatch: gcc

FantasqueX:
Fix rotten.

There are quite a lot of FAILs, however it seems that ArchLinux doesn't care about them.

```
make -O -k check || true
```

### DIFF
--- a/gcc/riscv64.patch
+++ b/gcc/riscv64.patch
@@ -1,17 +1,17 @@
-diff --git PKGBUILD PKGBUILD
-index 1952b31..37569f0 100644
---- PKGBUILD
-+++ PKGBUILD
+Index: PKGBUILD
+===================================================================
+--- PKGBUILD	(revision 455113)
++++ PKGBUILD	(working copy)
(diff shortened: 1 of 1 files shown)


FantasqueX:
BTW, glibc can update to 2.36-3 after this PR getting merged :)

gcc/riscv64.patch
```diff
@@ -26,24 +26,31 @@ index 1952b31..37569f0 100644
...
        --with-linker-hash-style=gnu
        --with-system-zlib
        --enable-__cxa_atexit
-@@ -94,7 +89,7 @@ build() {
+@@ -94,7 +89,6 @@
        --enable-link-serialization=1
        --enable-linker-build-id
        --enable-lto
 -      --enable-multilib
-+      --disable-multilib
```
r-value, line 42: unnecessary rearrange

FantasqueX: I just want to align enables and disables. It doesn't matter.

PR #1695, opened 2022-09-12
addpkg: python-regex

hilaolu:
modify `_pyver` local variable to pass check.


felixonmars:
Closing as the it's been addressed by https://github.com/felixonmars/archriscv-packages/issues/1732

PR #1696, opened 2022-09-12
addpkg: open-isns

hilaolu:
```
This script, last modified 2005-05-27, has failed to recognize
the operating system you are using. 
...
configure: error: cannot guess build type; you must specify one
```
The script can't detect the build type, thus we specify it manually by adding `--host=riscv64-unknown-linux-gnu` and  `--build=riscv64-unknown-linux-gnu` to `configure` parameters.

### DIFF
--- a/open-isns/riscv64.patch
+++ b/open-isns/riscv64.patch
@@ -1,0 +1,12 @@
+diff --git PKGBUILD PKGBUILD
+--- PKGBUILD
++++ PKGBUILD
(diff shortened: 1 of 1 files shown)


CoelacanthusHex:
Hi, the solution is in [our doc](https://riscv-notes.sh1mar.in/docs/record/collection#configguess), you should update config.guess by `autoreconf -fiv` and report to upstream.

PR #1699, opened 2022-09-13
addpkg: github-cli

hilaolu:
specify `GOARCH=riscv64` explicitly


r-value (changes requested):
Generally `GOARCH` shouldn't be set here.

Repros shows that there is probabilistic failures by entering an irrecoverable sleep under `qemu-user`, and an almost-certain success on RISC-V hardware. There should be a better fix.

In my opinion, marking the package as `flaky` or `noqemu` will be better.

PR #1717, opened 2022-09-19
addpkg: abletonlink

Ast-x64:


abletonlink/riscv64.patch
```diff
@@ -0,0 +1,13 @@
+diff --git PKGBUILD PKGBUILD
+index 392f00d1..6afed9d5 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -35,6 +35,8 @@ prepare() {
+ }
+ 
+ build() {
++  export CFLAGS="$CFLAGS -ffp-contract=off"
++  export CXXFLAGS="$CXXFLAGS -ffp-contract=off"
```
felixonmars, line 10: Can we have more info here? It's not immediately clear to me why `-ffp-contract` should be turned off.

felixonmars:
Superseded by #2708

PR #1721, opened 2022-09-20
addpatch: grafana-zabbix

Avimitin:
- Add dependency `python3`

Dependency node-gyp will download pre-built library in x86_64. In riscv64 it has to build itself from source which requires python3 executable. But there is no python package provided in the original PKGBUILD.

- Move the build script from Makefile to PKGBUILD

 Targets in Makefile doesn't handle all the platform. Maintainer hard code `amd64` suffix in `make build` and set environment `GOOS` to darwin, windows, openbsd...etc. It is not a proper way to run make build and make dist here.

Ref:
 - https://github.com/alexanderzobnin/grafana-zabbix/blob/master/Makefile#L45

Addtional:
  1. yarn install depends on network, so the make install command is flaky.
  2. Test will fail with signal `SIGSEGV` in QEMU-user. It is caused by test framework jest. This problem can be steadily reproduced.

### DIFF
--- a/grafana-zabbix/riscv64.patch
+++ b/grafana-zabbix/riscv64.patch
@@ -1,0 +1,23 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -7,15 +7,17 @@ arch=('any')
+ url="https://github.com/alexanderzobnin/grafana-zabbix"
+ license=('APACHE')
+ depends=('grafana')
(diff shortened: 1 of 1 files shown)


Avimitin:
Bug Tracker for the Makefile: https://bugs.archlinux.org/task/75150?project=5&string=grafana-zabbix

PR #1738, opened 2022-09-23
addpkg: ob-xd

Avimitin:
Replace the build option `-m64` to `-march=rv64gc`.

* Additional Notes

  This package can't be built in qemu-user because the build file generator Projucer will stuck.


felixonmars:
> * This package can't be built in qemu-user because the build file generator Projucer will stuck.

Please also add it to qemu-user-blacklist then.

Avimitin:
Changes are pushed.

PR #1752, opened 2022-09-24
addpkg: i7z to blacklist

bastianplettner:
The package i7z does not have hardware support for RISC-V.
It is a Intel/x86-specific CPU power state reporting tool.


felixonmars:
Makes sense. Thanks!

PR #1758, opened 2022-09-25
addpkg: exiv2

moodyhunter:
the `char` type on RISC-V is unsigned where the failed test assumed it to be signed[1]
apply a patch similar to what GUIX has done[2].

[1] https://github.com/Exiv2/exiv2/issues/933
[2] https://github.com/guix-mirror/guix/commit/b8df15b748024128e09cebb4d7ebae42be5df788


FantasqueX:
https://github.com/Exiv2/exiv2/issues/933#issuecomment-817712348

According to comment from maintainer, we are happy to accept tested PR which don't disturb our test suite running on the platforms we actively support.

Is it possible to send a PR to upstream?

moodyhunter:
The upstream `master` branch has changed too much, while the problem doesn't exist anymore on its latest revision. 

Probably we can merge this patch for now and remove it in the future once a new release comes.

PR #1759, opened 2022-09-25
qemu-user-blacklist: add ctags

moodyhunter:
cannot be built on qemu-user because several `stdout` and `exit` tests fail

### DIFF
--- a/qemu-user-blacklist.txt
+++ b/qemu-user-blacklist.txt
@@ -7,6 +7,7 @@ bmake
 caps
 cargo-audit
 cargo-nextest
+ctags
 coreutils
 datovka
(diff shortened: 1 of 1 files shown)


qemu-user-blacklist.txt
```diff
@@ -7,6 +7,7 @@ bmake
 caps
 cargo-audit
 cargo-nextest
+ctags
 coreutils
```
Xeonacid, line 11:
```suggestion
coreutils
ctags
```

Xeonacid (changes requested):

Xeonacid (approved):

PR #1761, opened 2022-09-26
addpkg: edac-utils

moodyhunter:
upstream's autoconf scripts are too old (from 2004), do `autoreconf` before compiling 

ref: https://github.com/grondo/edac-utils/pull/16


moodyhunter:
also changed `git://` to `git+https://`

PR #1763, opened 2022-09-26
addpkg: handbrake

bastianplettner:
intel-media-sdk is unavailable on riscv
The nvidia video codec sdk is unavailable on riscv (including nvenc)

basically the same as ffmpeg 057be4d0c5eadc6d1df953485180ea64387f93fc


r-value (approved):
Reproduced.

PR #1766, opened 2022-09-26
addpkg: cpupower to blacklist

bastianplettner:
cpupower is only compatible with x86

compare linux-tools patch 70c606f


Xeonacid:
I have reservations about this. cpupower seems not restricted to x86, which might work on riscv if firmware supports.

felixonmars:
Closing as `cpupower` is re-enabled in #1804

PR #1795, opened 2022-09-29
addpkg: libatomic_ops

yining765:
Update config.guess.
Issue reported: https://github.com/ivmai/libatomic_ops/issues/54

### DIFF
--- a/libatomic_ops/riscv64.patch
+++ b/libatomic_ops/riscv64.patch
@@ -1,0 +1,10 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -18,6 +18,7 @@
+ 
+ build() {
+   cd $pkgname-$pkgver
++  autoreconf -fi # Update config.guess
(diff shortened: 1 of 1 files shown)


libatomic_ops/riscv64.patch
```diff
@@ -0,0 +1,10 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -18,6 +18,7 @@
+ 
+ build() {
+   cd $pkgname-$pkgver
++  autoreconf -fi # Update config.guess
```
FantasqueX, line 7: We have a guide line about config.guess.

对于 `config.guess` 的问题，推荐改法：

```
prepare() {
    cd "${srcdir}/${_pkgname}-${pkgver}"
+   autoupdate      # this is optional, unless autoreconf complains
+   autoreconf -fiv
```

Please add verbose option.

FantasqueX (changes requested):
Please add verbose option according to our guide line.

PR #1827, opened 2022-10-10
Remove `rust` from qemu-user blacklist

r-value:
It's been a long time not observing unrecoverable sleep under qemu-user for packages written in Rust. Try to remove it.

### DIFF
--- a/qemu-user-makedepends-blacklist.txt
+++ b/qemu-user-makedepends-blacklist.txt
@@ -1,3 +1,1 @@
 go
-rust
-cargo


qemu-user-makedepends-blacklist.txt
```diff
@@ -1,3 +1,2 @@
 go
-rust
 cargo
```
felixonmars, line 2: cargo should be removed as well.

felixonmars (changes requested):

PR #1858, opened 2022-10-17
fixpkg: systemd

XieJiSS:
See also:

+ https://wiki.archlinux.org/title/Meson_package_guidelines#Using_arch-meson_wrapper_script
+ https://github.com/archlinux/svntogit-packages/blob/packages/meson/trunk/arch-meson
+ https://github.com/systemd/systemd/commit/55b5daf9b2b7e0c8b77de4f87986832447e72b84 (not appliable because of missing source file; waiting for this to be bpo into `systemd-stable` before we can enable LTO)

Sonames differ in systemd.

### DIFF
--- a/systemd/riscv64.patch
+++ b/systemd/riscv64.patch
@@ -9,16 +9,49 @@
  options=('debug' 'strip')
  validpgpkeys=('63CDA1E5D3FC22B998D20DD6327F26951A015CC4'  # Lennart Poettering <lennart@poettering.net>
                'A9EA9081724FFAE0484C35A1A81CEA22BC8C7E2E'  # Luca Boccassi <luca.boccassi@gmail.com>
-@@ -149,7 +149,7 @@ build() {
+@@ -68,6 +68,13 @@ _backports=(
+ _reverts=(
+ )
+ 
++
++source+=("allow-riscv-flush-icache.patch::https://github.com/systemd/systemd/commit/09925036cf2b5a5c4cf680422a38c427ca692cd6.diff")
++
++sha512sums+=('06310189883f3881585804619ee475ebac914f7563a8e1776b6f4abb56cabc8e1e3ce805514270810262e72e60538db2354e2e8be7de68d665fc002a7af77626')
++
++options+=(!lto)
++
+ prepare() {
+   cd "$pkgbase-stable"
+ 
+@@ -90,6 +97,8 @@ prepare() {
+ 
+   # Replace cdrom/dialout/tape groups with optical/uucp/storage
+   patch -Np1 -i ../0001-Use-Arch-Linux-device-access-groups.patch
++
++  patch -Np1 -i ../allow-riscv-flush-icache.patch
  }
-
+ 
+ build() {
+@@ -113,6 +122,9 @@ build() {
+     -Dversion-tag="${_tag_name/-/\~}-${pkgrel}-arch"
+     -Dshared-lib-tag="${pkgver}-${pkgrel}"
+     -Dmode=release
++    # disable b_lto which is enabled in arch-meson
++    # https://github.com/systemd/systemd/issues/24630, but patch file does not match
++    -D b_lto=false
+ 
+     -Dgnu-efi=true
+     -Dima=false
+@@ -151,7 +163,7 @@ build() {
+ }
+ 
  check() {
 -  meson test -C build
 +  meson test -C build -t 10
  }
-
+ 
  package_systemd() {
-@@ -169,8 +169,7 @@ package_systemd() {
+@@ -171,8 +183,7 @@ package_systemd() {
                'systemd-sysvcompat: symlink package to provide sysvinit binaries'
                'polkit: allow administration as unprivileged user'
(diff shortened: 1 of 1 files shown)


systemd/riscv64.patch
```diff
@@ -9,16 +9,59 @@
...
+@@ -147,13 +156,24 @@ build() {
+     -Dsbat-distro-url="https://archlinux.org/packages/core/x86_64/${pkgname}/"
+   )
+ 
+-  arch-meson "$pkgbase-stable" build "${_meson_options[@]}"
+-
+-  meson compile -C build
++  # disable b_lto which is enabled in arch-meson
++  # https://github.com/systemd/systemd/issues/24630, but patch file does not match
++  meson setup \
```
CoelacanthusHex, line 44: Just add `-Db_lto=false` as argument of `arch-meson`. Meson will use last one if conflict.

XieJiSS: > Just add `-Db_lto=false` as argument of `arch-meson`. Meson will use last one if conflict.

Tried before (w/ or w/o options=(!lto)), seems like it does not take effect

XieJiSS: Can confirm that using `arch-meson` with `-D b_lto=false` will result in `-flto` being appended to the CC args. Note that this is not caused by `options=(lto)`, which generates `-flto=auto` instead of `-flto`.

```
/usr/bin/cc -o src/boot/efi/systemd-bootriscv64.elf -Wno-format-signedness -Wno-missing-field-initializers -Wno-unused-parameter -Wdate-time -Wendif-labels -Werror=format=2 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=overflow -Werror=override-init -Werror=return-type -Werror=shift-count-overflow -Werror=shift-overflow=2 -Werror=undef -Wfloat-equal -Wimplicit-fallthrough=5 -Winit-self -Wlogical-op -Wmissing-include-dirs -Wmissing-noreturn -Wnested-externs -Wold-style-definition -Wpointer-arith -Wredundant-decls -Wshadow -Wstrict-aliasing=2 -Wstrict-prototypes -Wsuggest-attribute=noreturn -Wunused-function -Wwrite-strings -Wno-maybe-uninitialized -Wno-unused-result -fno-stack-protector -fno-strict-aliasing -fpic -fwide-exec-charset=UCS2 -Wall -Wextra -Wsign-compare -nostdlib -std=gnu11 -ffreestanding -fshort-wchar -fvisibility=hidden -isystem /usr/include/efi -isystem /usr/include/efi/riscv64 -I /build/systemd/src/systemd-stable/src/fundamental -DSD_BOOT -DGNU_EFI_USE_MS_ABI -include src/boot/efi/efi_config.h -include version.h -O1 -flto -O2 -ffile-prefix-map=/build/systemd/src=/usr/src/debug -fuse-ld=bfd -L /usr/lib -nostdlib -T /usr/lib/elf_riscv64_efi.lds -Wl,--build-id=sha1 -Wl,--fatal-warnings -Wl,--no-undefined -Wl,--warn-common -Wl,-Bsymbolic -z nocombreloc /usr/lib/crt0-efi-riscv64.o -Wl,--no-warn-execstack -Wl,--no-warn-rwx-segments -shared -Wl,--defsym=EFI_SUBSYSTEM=0xa src/boot/efi/bootspec-fundamental.c.o src/boot/efi/efivars-fundamental.c.o src/boot/efi/sha256.c.o src/boot/efi/string-util-fundamental.c.o src/boot/efi/assert.c.o src/boot/efi/devicetree.c.o src/boot/efi/disk.c.o src/boot/efi/graphics.c.o src/boot/efi/measure.c.o src/boot/efi/pe.c.o src/boot/efi/secure-boot.c.o src/boot/efi/ticks.c.o src/boot/efi/util.c.o src/boot/efi/boot.c.o src/boot/efi/console.c.o src/boot/efi/drivers.c.o src/boot/efi/random-seed.c.o src/boot/efi/shim.c.o src/boot/efi/xbootldr.c.o -lefi -lgnuefi -lgcc
/usr/bin/ld.bfd: /tmp/ccwqtUMf.ltrans0.ltrans.o: in function `entries_unique':
/usr/src/debug/build/../systemd-stable/src/boot/efi/boot.c:1793: undefined reference to `memset'
/usr/bin/ld.bfd: /tmp/ccwqtUMf.ltrans0.ltrans.o: in function `.L0 ':
/build/gnu-efi/src/gnu-efi-3.0.15//lib/misc.c:34: undefined reference to `memset'
/usr/bin/ld.bfd: /tmp/ccwqtUMf.ltrans0.ltrans.o: in function `.L0 ':
/usr/src/debug/build/../systemd-stable/src/boot/efi/boot.c:1171: undefined reference to `memset'
/usr/bin/ld.bfd: /tmp/ccwqtUMf.ltrans0.ltrans.o: in function `.L0 ':
/build/gnu-efi/src/gnu-efi-3.0.15//lib/str.c:232: undefined reference to `memset'
collect2: error: ld returned 1 exit status
```

PR #1867, opened 2022-10-22
rmpatch: ansible-lint

FantasqueX:
Upstream fixed in https://github.com/archlinux/svntogit-community/commit/cb5d2f7bae0a82a8bb37226e8e0769a0f579916b

I think I shouldn't send a PR to fix validity next time, just waiting for upstream fix is fine.


felixonmars:
> I think I shouldn't send a PR to fix validity next time, just waiting for upstream fix is fine.

A better approach is to give each step some time, like sending the patch here when Arch upstream didn't respond in a week or so.

Actually, no worries :)

PR #1868, opened 2022-10-24
upgpatch: linux 6.0.2.arch1-1

Xeonacid:
1. Backport [RISC-V: Make mmap() with PROT_WRITE imply PROT_READ](https://git.kernel.org/pub/scm/linux/kernel/git/riscv/linux.git/commit/?h=for-next&id=8aeb7b17f04ef40f620c763502e2b644c5c73efd).
2. Force disable sv57, since many packages ]like go and openjdk cannot handle it.
3. Disable VMAP_STACK, or the kernel will crash on Unmatched with UEFI. [Ubuntu bug report](https://bugs.launchpad.net/ubuntu/+source/linux-riscv/+bug/1967130)

Note for reviewers: [Arch's `prepare()`](https://github.com/archlinux/svntogit-packages/blob/packages/linux/repos/core-x86_64/PKGBUILD#L43-L50) will apply all patches with 	`.patch` extension, so we don't need to patch manually.


linux/riscv64.config-patch
```diff
@@ -54,15 +54,15 @@
  # end of RISC-V CPU Idle Drivers
  # end of CPU Idle
  # end of CPU Power Management
-@@ -1737,6 +1744,7 @@
- CONFIG_PCI_STUB=y
- CONFIG_PCI_PF_STUB=m
- CONFIG_PCI_ATS=y
-+CONFIG_PCI_ECAM=y
```
XieJiSS, line 61: Is the removal of `+CONFIG_PCI_ECAM=y` intended?

linux/riscv64.config-patch
```diff
@@ -206,17 +206,7 @@
  CONFIG_DMADEVICES=y
  # CONFIG_DMADEVICES_DEBUG is not set
  
-@@ -7822,7 +7846,8 @@
- #
- # VME Device Drivers
- #
--# CONFIG_GOLDFISH is not set
```
XieJiSS, line 213: ditto

linux/riscv64.config-patch
```diff
@@ -206,17 +206,7 @@
...
-+CONFIG_GOLDFISH=y
-+# CONFIG_GOLDFISH_PIPE is not set
```
XieJiSS, line 215: ditto

linux/riscv64.config-patch
```diff
@@ -206,17 +206,7 @@
  CONFIG_DMADEVICES=y
  # CONFIG_DMADEVICES_DEBUG is not set
  
-@@ -7822,7 +7846,8 @@
- #
- # VME Device Drivers
- #
--# CONFIG_GOLDFISH is not set
-+CONFIG_GOLDFISH=y
```
XieJiSS, line 214: ditto

linux/riscv64.config-patch
Xeonacid: This is modified by `make olddefconfig` and seems almost unrelated to riscv.

XieJiSS (approved):
LGTM

PR #1869, opened 2022-10-24
upgpatch: java-openjdk 19.0.1.u10-2

Xeonacid:
We (nearly) don't need any patch for Java 19 🎉 
Should work on kernel built with #1868 (with sv57 patch).


XieJiSS (approved):
Congrats

PR #1889, opened 2022-10-27
updpatch: gcc

xctan:
Add riscv support to libitm and enable it in PKGBUILD


xctan:
upstreamed: https://gcc.gnu.org/pipermail/gcc-patches/2022-October/604477.html

PR #1894, opened 2022-10-29
addpkg: renpy

Ast-x64:


felixonmars:
Perhaps worth a bug report to change it to $CARCH. It's a namcap warning as well :)

PR #1929, opened 2022-11-09
addpatch: limesuite

FantasqueX:
LimeSuite enables `-march=native` by default. This patch adds a check. If compiler supports `-march=native`, we can enable it by default. This patch has been merged upstream.
Upstream link: https://github.com/myriadrf/LimeSuite/pull/370


felixonmars:
This is also a bug in Arch x86_64 as we should not use `-march=native` in packaging there, either. Please also report it to Arch :)

FantasqueX:
Reported to ArchLinux here https://bugs.archlinux.org/task/76498

PR #1944, opened 2022-11-14
addpatch: clamav

Avimitin:
The rust test need more time to finish on our machine.

### DIFF
--- a/clamav/riscv64.patch
+++ b/clamav/riscv64.patch
@@ -1,0 +1,13 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -66,8 +66,8 @@ build() {
+ }
+ 
+ check() {
+-  cd ${pkgname}-${pkgver}
+-  ninja -C build test
++  cd ${pkgname}-${pkgver}/build
++  ctest --timeout 5000
+ }
+ 
+ package() {


clamav/riscv64.patch
```diff
@@ -0,0 +1,13 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -66,8 +66,8 @@ build() {
+ }
+ 
+ check() {
+-  cd ${pkgname}-${pkgver}
+-  ninja -C build test
++  cd ${pkgname}-${pkgver}/build
++  ctest --timeout 5000
```
Xeonacid, line 10: Why dropped ninja here?

Avimitin: `ninja -C build test` and `ctest` are doing the same thing. `ninja test` just run the `test` target and invoke the ctest executable. And there is no way to pass the timeout argument down to the ctest executable. It will be easier to configure test timeout value when we just use ctest.

Avimitin: And ninja team explain why they don't pass the argument to target in this issue: https://github.com/ninja-build/ninja/issues/1387#issuecomment-449394673

Xeonacid (approved):

PR #1957, opened 2022-11-17
addpatch: glpk

Avimitin:
Fix config.guess cannot guess build type issue.


Avimitin:
Reported through [bug-glpk@gnu.org](mailto:bug-glpk@gnu.org).

PR #1977, opened 2022-11-21
addpatch: dateutils

xctan:
Fix config.guess


felixonmars:
Don't forget to notify upstream abou this :)

PR #1991, opened 2022-11-24
addpatch: partimage

xctan:
Fix config.guess


felixonmars:
Please also contact upstream to update their release tarballs :)

PR #2017, opened 2022-11-28
addpatch: coin-or-cgl

Avimitin:
Fix the config.guess issue. BTW, the coin-or-clp in current repo link
to the libcholmod.so.3 file, but the latest suitesparse package provide
libcholmod.so.4 now. So the coin-or-clp package should be built before
this package.


felixonmars:
The patch file is located in the wrong folder.

PR #2044, opened 2022-12-08
fixpatch: python-sphinx_rtd_theme

Avimitin:
Fix rotten and webpack4 issue.

The latest Node.js contains some breaking change for webpack4. This PR
add env `NODE_OPTIONS=--openssl-legacy-provider` as a workaround for the
`Error: error:0308010C:digital envelope routines::unsupported` issue.


Avimitin:
This issue only occurs on Arch RISC-V because the Arch upstream is using the old nodejs.

PR #2049, opened 2022-12-12
updpkg: onefetch

TinySnow:
Fixed rotten.
The maintainer's info has been changed, git diff contains it.
If that should not be included, please tell me.


felixonmars:
> The maintainer's info has been changed, git diff contains it.
> If that should not be included, please tell me.

I actually didn't get you about "git diff contains it", but our patch definitely shouldn't change them.

PR #2094, opened 2023-01-04
addpkg: memtest86+ to blacklist

Avimitin:
As described on their homepage, this program is a memory tester for
x86 and x86-64 architecture computers.

### DIFF
--- a/blacklist.txt
+++ b/blacklist.txt
@@ -34,6 +34,7 @@ intel-ucode
 intel-undervolt
 libmfx
 libva-intel-driver
+memtest86+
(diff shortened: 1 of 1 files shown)


felixonmars:
Upstream is being conservative about adding more architectures to support. However, this doesn't stop anyone from working on a maintained fork, etc.

I would still want our blacklist to just filter out those we don't want to think about (closed-source stuff or hardware-specific stuff), instead of everything that upstream doesn't currently support.

Avimitin:
I get it.

PR #2102, opened 2023-01-09
addpkg: rust-rage

light4:
[pprof-rs](https://github.com/tikv/pprof-rs) master branch add support for riscv64, but not released yet

### DIFF
--- a/rust-rage/riscv64.patch
+++ b/rust-rage/riscv64.patch
@@ -1,0 +1,24 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -13,12 +13,18 @@ depends=(bzip2
+          fuse2)
+ makedepends=(cargo)
+ _archive="$_pkgname-$pkgver"
+-source=("$url/archive/v$pkgver/$_archive.tar.gz")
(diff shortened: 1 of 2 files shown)


rust-rage/riscv64.patch
```diff
@@ -0,0 +1,24 @@
...
++sha256sums=('f5756004061f974e79607224864ceb3fdd50367272cc58aa5a755929cf1c70a3'
++            '3c38c189646bdb712527a3ae4118dc72c5746576b1e90510e914342a0225de03')
+ 
+ prepare() {
+ 	cd "$_archive"
+-	cargo fetch --locked --target "$CARCH-unknown-linux-gnu"
++
++	patch -Np1 -i ../fix_rage_deps.patch # pprof-rs add riscv64 support, but not published to crates.io
++
++	cargo fetch
```
felixonmars, line 21: We should probably add `cargo update -p pprof` here instead of removing `--locked` from `cargo fetch` to avoid bumping other dependencies too.

felixonmars (changes requested):

PR #2104, opened 2023-01-10
upgpkg: cargo-msrv

light4:
original PKGBUILD set rust toolchain, fix patch

### DIFF
--- a/cargo-msrv/riscv64.patch
+++ b/cargo-msrv/riscv64.patch
@@ -1,9 +1,9 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -15,7 +15,9 @@ options=('!lto')
- 
+@@ -16,7 +16,9 @@ options=('!lto')
  prepare() {
    cd "$pkgname-$pkgver"
+   export RUSTUP_TOOLCHAIN=stable
 -  cargo fetch --locked --target "$CARCH-unknown-linux-gnu"
 +  echo -e "\n[patch.crates-io]\nring = { git = 'https://github.com/felixonmars/ring', branch = '0.16.20' }" >> Cargo.toml
 +  cargo update -p ring


SpriteOvO:
@light4 Unfortunately, this package still does not pass the tests.

```
failures:
    minimum_from_edition::msrv_min_with_edition_in_cargo_toml
    minimum_from_edition::msrv_no_minimum_with_flag
    msrv_unsupported
    msrv_using_bisect_method::case_0
    msrv_using_bisect_method::case_1
    msrv_using_bisect_method::case_2
    msrv_using_bisect_method::case_3
    msrv_using_linear_method::case_0
    msrv_using_linear_method::case_1
    msrv_using_linear_method::case_2
    msrv_using_linear_method::case_3
    msrv_with_custom_command::case_0
    msrv_with_custom_command::case_1
    msrv_with_custom_command::case_2
    msrv_with_custom_command::case_3
    msrv_with_old_lockfile
    msrv_with_release_source::case_0
    msrv_with_release_source::case_1

test result: FAILED. 6 passed; 18 failed; 0 ignored; 0 measured; 0 filtered out; finished in 139.88s
```

The main challenge here is that Rust does not officially provide binaries with `riscv64gc` triples, and the tests of this package requires installing old versions of Rust by `rustup`.

SpriteOvO:
> Maybe disable some tests on riscv? I'll make a PR to upstream.

@light4 I prefer to leave this package broken, as:

- The problem is on `rustup`.

- It is a bin package, it does not block other packages as a dependency, so disabling tests does not provide benefits.

- Even if the tests are disabled, the package will not work at all for users.

light4:
This package actually works fine at riscv64. The dist `riscv64gc` became available from [1.47.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1470-2020-10-08) 
Disable some little tests won't affect users.

SpriteOvO:
> This package actually works fine at riscv64. The dist `riscv64gc` became available from [1.47.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md?rgh-link-date=2023-01-10T15%3A01%3A23Z#version-1470-2020-10-08) Disable some little tests won't affect users.

@light4 Oh, you are right, so this is actually a bug from package upstream. It would be great if you would open a PR fix it there :)

r-value:
I suggest fixing the package by adding a global MSRV for certain platforms or file an issue like "Panicked on new platforms if MSRV is too old" instead of just disabling the test at upstream. If you fixed it, you can take the patch here before it's merged and released to accelerate the fixing process.

light4:
Filed an issue https://github.com/foresterre/cargo-msrv/issues/585.

PR #2105, opened 2023-01-10
addpkg: cargo-tarpaulin

light4:
fix rust target


r-value (commented):
Generally we recommend just removing the `--target` parameter for Rust target triple issues.

light4:
updated

PR #2111, opened 2023-01-13
addpkg: puzzles

a1ca7raz:


r-value:
Thank you for your contribution! But it seems that this patch is an update ahead of Arch Linux. Generally we prefer following the upstream updates. Maybe you can flag the package as out-of-data at [Arch Linux](https://archlinux.org/packages/community/x86_64/puzzles/)?

PR #2117, opened 2023-01-15
addpkg: yasm

a1ca7raz:


yasm/rm_ac_cv_header_stdc.patch
```diff
@@ -0,0 +1,14 @@
...
++++ configure.ac	2023-01-15 10:40:31.900283925 +0800
+@@ -112,11 +112,6 @@ AC_PROG_EGREP
+ 
+ AC_CHECK_HEADERS([strings.h libgen.h unistd.h direct.h sys/stat.h])
+ 
+-# REQUIRE standard C headers
+-if test "$ac_cv_header_stdc" != yes; then
+-	AC_MSG_ERROR([Standard (ANSI/ISO C89) header files are required.])
+-fi
+-
```
felixonmars, line 11: We generally want to avoid maintaining downstream patches like this. Please try to figure out the reason and send an appropriate patch to upstream first.

PR #2121, opened 2023-01-16
upgpkg: devtools-riscv64 use patch instead of fork

Xeonacid:
Sync with devtools first. The way to trigger rebuild automatically will be updated later.

Changes:
- `makepkg-riscv64.conf`:
    - Add back `-fno-plt`
- `pacman-extra-riscv64.conf`:
    - Use `Architecture = auto` instead of `Architecture = riscv64`
    - ~~Use `pacman-mirrorlist` for repos~~


felixonmars:
> * Use `pacman-mirrorlist` for repos

This would break devtools because it lacks proper handling of mirrorlist now: https://gitlab.archlinux.org/archlinux/devtools/-/issues/104

Xeonacid:
> > * Use `pacman-mirrorlist` for repos
> 
> This would break devtools because it lacks proper handling of mirrorlist now: [gitlab.archlinux.org/archlinux/devtools/-/issues/104](https://gitlab.archlinux.org/archlinux/devtools/-/issues/104)

Fixed

PR #2149, opened 2023-01-23
addpkg: aarch64-linux-gnu-binutils

a1ca7raz:
$pkgdir/usr/etc/gprofng.rc does not exist on riscv64 platform.

### DIFF
--- a/aarch64-linux-gnu-binutils/riscv64.patch
+++ b/aarch64-linux-gnu-binutils/riscv64.patch
@@ -1,0 +1,12 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -61,9 +61,6 @@ package() {
+   # Remove file conflicting with host binutils and manpages for MS Windows tools
+   rm "$pkgdir"/usr/share/man/man1/$_target-{dlltool,windres,windmc}*
+   rm "$pkgdir"/usr/lib/bfd-plugins/libdep.so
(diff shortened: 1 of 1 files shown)


Xeonacid (approved):

felixonmars:
Any idea why they aren't present? It would make sense to investigate so we understand the patch better in the future.

xctan:
> Any idea why they aren't present? It would make sense to investigate so we understand the patch better in the future.

Looks like these files related to gprofng aren't configured to be placed at proper location. The configure.ac file [only covers](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=configure.ac;h=e95a9ed116e83f43e741e9a691b15fc194d23d4a;hb=HEAD#l414) x86, x86_64, and aarch64 platform.

felixonmars:
> > Any idea why they aren't present? It would make sense to investigate so we understand the patch better in the future.
> 
> Looks like these files related to gprofng aren't configured to be placed at proper location. The configure.ac file [only covers](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=configure.ac;h=e95a9ed116e83f43e741e9a691b15fc194d23d4a;hb=HEAD#l414) x86, x86_64, and aarch64 platform.

Thanks! It looks more like gprofng has no support for other architectures to me though. Either way, we could apply this patch safely now.

PR #2156, opened 2023-01-26
updpatch: mesa

FantasqueX:
This patch only disables mesa intel.

### DIFF
--- a/mesa/riscv64.patch
+++ b/mesa/riscv64.patch
@@ -1,44 +1,80 @@
 Index: PKGBUILD
 ===================================================================
---- PKGBUILD	(revision 444265)
+--- PKGBUILD	(revision 467369)
 +++ PKGBUILD	(working copy)
-@@ -6,3 +6,3 @@
- pkgbase=mesa
--pkgname=('vulkan-mesa-layers' 'opencl-mesa' 'vulkan-intel' 'vulkan-radeon' 'vulkan-swrast' 'libva-mesa-driver' 'mesa-vdpau' 'mesa')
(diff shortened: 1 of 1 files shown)


mesa/riscv64.patch
```diff
@@ -1,44 +1,45 @@
...
      -D dri3=enabled \
-@@ -53,3 +54,3 @@
+     -D egl=enabled \
      -D gallium-extra-hud=true \
--    -D gallium-nine=true \
-+    -D gallium-nine=false \
-     -D gallium-omx=bellagio \
-@@ -67,2 +68,3 @@
-     -D llvm=enabled \
-+    -D draw-use-llvm=false \
```
felixonmars, line 39: draw-use-llvm was disabled because llvmpipe doesn't work properly. Have you tested that it works correctly now without the orcjit patch?

FantasqueX: Sorry for my carelessness. Since I don't have a headful device currently, I'll disable llvmpipe later.

felixonmars: An easier test case could be just running `ssh -Y` or even xvfb for glxgears etc. Previously it just crashes with some LLVM errors.

FantasqueX:
Hi, I've updated the branch.

This patch disables all intel drivers and disables draw-use-llvm because
llvmpipe doesn't work properly on riscv64. This patch also disables
swrast vulkan driver because Lavapipe requires LLVM draw support if LLVM
is enabled.

Changes from last patch:
1. Readd directx-headers as makedepends and install d3d library.
2. Reenable gallium-nine because we cannot build swrast vulkan driver instead
of swrast gallium driver.

PR #2172, opened 2023-02-04
addpkg: m4 to qemu-user-blacklist

bastianplettner:
Some tests fail in qemu-user but pass just fine on riscv hardware.

```
============================================================================
Testsuite summary for GNU M4 1.4.19
============================================================================
# TOTAL: 267
# PASS:  234
# SKIP:  33
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

### DIFF
--- a/qemu-user-blacklist.txt
+++ b/qemu-user-blacklist.txt
@@ -52,6 +52,7 @@ liborcus
 libseccomp
 libsecret
 libuv
+m4
 mdbook
 meilisearch
 memcached


FantasqueX:
Hi, do you know why those tests fail? Some tests pass on real hardware just because performance of those hardware is bad such as gnutls. The summary of successful tests is useless while the summary of failed tests running on qemu-user is more important. Afterwards packagers will use the information of those failed tests. So, could you please also add the summary of failed tests?

bastianplettner:
The failing tests are:
```
stackovf.test
test-c-stack.sh
test-sigsegv-catch-stackoverflow1
test-sigsegv-catch-stackoverflow2
```
All four seem to be related to testing for stackoverflow handling.

PR #2184, opened 2023-02-10
upgpatch: cargo-spellcheck

Avimitin:
cargo-spellcheck introduce new dependency perf-event-open-sys which
doesn't have riscv64 binding. This commit change its source to my
patched version.


Avimitin:
Those riscv64 implementation are submitted to the upstream: https://github.com/jimblandy/perf-event/pull/28

PR #2194, opened 2023-02-15
updpatch: qemu 7.2.0-3

bastianplettner:
fix patch not applying to qemu 7.2.0-3

### DIFF
--- a/qemu/riscv64.patch
+++ b/qemu/riscv64.patch
@@ -1,31 +1,30 @@
-Index: PKGBUILD
-===================================================================
---- PKGBUILD	(revision 459941)
-+++ PKGBUILD	(working copy)
-@@ -97,7 +97,7 @@
+diff --git PKGBUILD PKGBUILD
+index 768b1eb..500bd3d 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -97,6 +97,7 @@ makedepends=(
    zlib
    zstd
  )
--options=(debug)
-+options=(debug !lto)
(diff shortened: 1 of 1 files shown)


qemu/riscv64.patch
```diff
@@ -59,16 +58,16 @@ Index: PKGBUILD
...
      _pick qemu-system-xtensa usr/bin/qemu-system-xtensa{,eb}
      _pick qemu-system-xtensa usr/share/man/man1/qemu-system-xtensa{,eb}.1*
-@@ -716,7 +722,7 @@
+@@ -711,7 +718,7 @@ package_qemu-hw-s390x-virtio-gpu-ccw() {
  
  package_qemu-system-aarch64() {
    pkgdesc="QEMU system emulator for AARCH64"
--  depends=("${_qemu_system_deps[@]}" edk2-armvirt systemd-libs libudev.so)
+-  depends=("${_qemu_system_deps[@]}" edk2-aarch64 systemd-libs libudev.so)
 +  depends=("${_qemu_system_deps[@]}" systemd-libs libudev.so)
```
felixonmars, line 66: The `edk2-arm` dependency of `qemu-system-arm` package needs to be removed as well, or the package would become uninstallable.

felixonmars (changes requested):

qemu/riscv64.patch
bastianplettner: Sure, thanks for the input.

PR #2197, opened 2023-02-18
rmvpatch: ccls

TinySnow:
This package can be compiled without our patch now.
I observed that the version of clang depend in latest PKGBUILD changed to 15.0.7.
Maybe this is reason.

Build process tails:

```bash
./../bin/ccls
==> Running checkpkg
error: failed to prepare transaction (package architecture is not valid)
==> WARNING: Skipped checkpkg due to missing repo packages
```


Xeonacid:
> ```shell
> ./../bin/ccls
> ==> Running checkpkg
> error: failed to prepare transaction (package architecture is not valid)
> ==> WARNING: Skipped checkpkg due to missing repo packages
> ```

This should not anyhow be related to clang's version bump IMO.

TinySnow:
> This should not anyhow be related to clang's version bump IMO.

I just put it as testimony of building success, it's my phrase mistakes.
I don't know if the failure of preparing transaction will cause any problems.

PR #2229, opened 2023-02-26
updpatch: openblas

xctan:
Fix rotten patch, and fix cmake.arch (upstream pr: https://github.com/xianyi/OpenBLAS/pull/3923)


felixonmars:
Actually the `openblas` variant is not a 32-bit lib, it's a 64-bit lib with 32-bit integer (which is the default). The `openblas64` variant is specifically made for julia.

PR #2239, opened 2023-02-27
upgpatch: ldc

Avimitin:
ldc released initial riscv support in the latest version. This PR
remove all the old patches.

Notice: this package require extra `llvm14` dependency for rebuild.


Xeonacid:
@Avimitin I guess llvm14-libs is enough.

r-value:
Any updates?
Currently `ldc` in package repo is broken, which blocks test of `meson`, which blocks `mesa`.

PR #2254, opened 2023-02-28
addpatch: gimp-help

r-value:
Fix `config.guess`


Xeonacid:
Upstream is using an updated `config.guess`, closing.

PR #2263, opened 2023-03-02
addpatch: transmission

Avimitin:
The original PKGBUILD link `transmission` to the system b64 library.
However, the system b64 library did not handle the data type correctly
and broke the base64 test on riscv64. This commit removes the libb64
dependency and links transmission to its forked version of the b64
library.


felixonmars:
Asking Arch to switch libb64 to the fork (https://github.com/libb64/libb64) is also possible it seems.

PR #2285, opened 2023-03-09
qemu-user-blacklist: add lmdb

kxxt:
lmdb can be built in qemu-system but it fails to build in qemu-user because of an "Operation not supported" error raised in `check()`

Related: #2226


kxxt:
Superseded by #2835

PR #2289, opened 2023-03-10
feat(linux-config-maintain): read kernel options from file 

saeziae:


patch-linux-config.sh
```diff
@@ -5,18 +5,17 @@ if [[ ! -x ./scripts/config ]]; then
...
 
-configs=(
-    '-e' 'SOC_STARFIVE' # StarFive board support
-    '-e' 'SOC_SIFIVE' '-e' 'GPIO_SIFIVE' '-e' 'SIFIVE_CCACHE' '-e' 'EDAC_SIFIVE' '-e' 'PCIE_FU740' '-m' 'PWM_SIFIVE' # SiFive board support
-    '-e' 'ERRATA_THEAD' '-e' 'ERRATA_THEAD_PBMT' # # THead board support
-    '-e' 'SOC_MICROCHIP_POLARFIRE' '-e' 'PCIE_MICROCHIP_HOST' '-e' 'PCIE_XILINX' '-e' 'SERIAL_OF_PLATFORM' '-m' 'USB_MUSB_POLARFIRE_SOC' '-m' 'RTC_DRV_POLARFIRE_SOC' '-m' 'POLARFIRE_SOC_MAILBOX' '-m' 'I2C_MICROCHIP_CORE' '-m' 'POLARFIRE_SOC_SYS_CTRL' '-m' 'HW_RANDOM_POLARFIRE_SOC' # PolarFire board support
-    '-e' 'SOC_VIRT' '-e' 'PCI_HOST_GENERIC' # QEMU support
-    '-m' 'MMC_DW_PLTFM' '-m' 'MMC_DW'
-)
+if [[ ! -x "$1" ]]; then
```
CoelacanthusHex, line 8:
```suggestion
if [[ ! -r "$1" ]]; then
```
Should be a readable file rather than a executable file?

patch-linux-config.sh
```diff
@@ -5,18 +5,17 @@ if [[ ! -x ./scripts/config ]]; then
...
+  echo "No defconfig file given!"
+    exit 1
+fi
+
+configs=$(cat "$1" | grep -v '^#' | sed 's/^CONFIG_//' |
```
CoelacanthusHex, line 13:
```suggestion
configs=$(grep -v '^#' "$1" | sed 's/^CONFIG_//' |
```
Just use grep directly.

patch-linux-config.sh
saeziae: oops, my fault

CoelacanthusHex:
Please rebase.

CoelacanthusHex (approved):
LGTM

XieJiSS:
We need to specify `/g` flag for `sed`. Also I would like to suggest trimming before `awk` so that trailing whitespaces won't affect matching.

saeziae:
shell works by lines so i reckon `/g` is not needed.
 
trimming sounds useful.

PR #2291, opened 2023-03-12
addpkg: i2pd

hack3ric:


hack3ric:
Upstream flagged as outdated.

Xeonacid:
> Upstream flagged as outdated.

In such situation, we should file a bug report named like "checksum mismatch", instead of flagging OOD. :)

PR #2298, opened 2023-03-13
upgpatch: xplr 0.20.2-1

hack3ric:
Uses system `luajit` instead of vendored or `lua51`.


felixonmars:
This also worth a try to upstream to Arch, even though it requires patching x)

PR #2299, opened 2023-03-13
qemu-user-blacklist: add lmdb

xctan:
check() failed on qemu-user (operation not supported).


xctan:
dup of https://github.com/felixonmars/archriscv-packages/pull/2285

PR #2306, opened 2023-03-13
addpatch: menu-cache

kxxt:
Upstream hasn't been updated for 6 years so I guess there is not need to report to upstream.


XieJiSS:
If the upstream bug tracker is publicly available, your report can help others who encounter the same issue in the future

kxxt:
> If the upstream bug tracker is publicly available, your report can help others who encounter the same issue in the future

Thanks for your advice. I have opened a bug ticket on their bug tracker: https://sourceforge.net/p/lxde/bugs/962/

PR #2310, opened 2023-03-14
upgpatch: meson 1.0.1-1

hack3ric:
Fix rotten patch.


hack3ric:
Can confirm test is stuck in qemu-user. May try it again on qemu-system or real hardware.

PR #2350, opened 2023-03-18
upgpkg: python-ipykernel-bootstrap

kxxt:


hack3ric:
The package name should be `python-ipykernel-bootstrap`, not `python-ipython-bootstrap`.

kxxt:
> The package name should be `python-ipykernel-bootstrap`, not `python-ipython-bootstrap`.

My fault. It has been fixed now.

Xeonacid:
We don't need to upgrade the -bootstrap postfixed package, which is for bootstrap only, and the PKGBUILD is just an archive here.

PR #2352, opened 2023-03-18
updpatch: mold

xctan:
Disabled a new test (z-dynamic-undefined-weak, clang will hang) and disable MOLD_USE_MOLD (gcc omits default library path).

### DIFF
--- a/mold/riscv64.patch
+++ b/mold/riscv64.patch
@@ -1,18 +1,24 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -24,6 +24,11 @@ pkgver() {
+@@ -23,6 +23,10 @@ pkgver() {
  }
  
  build() {
-+  for failing_test in lto-gcc gnu-unique package-metadata tls-common weak-undef-dso riscv-norvc tail-call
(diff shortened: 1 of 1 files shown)


felixonmars:
I actually didn't get the last part. Is disabling MOLD_USE_MOLD a temporary workaround due to our mold isn't usable as-is at the moment (linking libLLVM-14.so), or probably something worth mentioning to mold upstream?

xctan:
We are using our gcc to build mold, but unfortunately our gcc doesn't pass "default" lib paths like /usr/lib to the linker. This is fine with bfd, which automatically searches conventional paths, but mold cannot work without this information. Maybe there is a problem with our gcc spec regarding this issue.

Two approaches are taken to workaround the problem:

* Use clang with mold, which correctly invokes the linker. The tests are forced to use clang and clang++, but because of incompatible command line switches with gcc, some tests must be disabled;
* Use gcc with bfd instead of mold, so MOLD_USE_MOLD is disabled when building mold.

felixonmars:
I see. That's unfortunate :(

Sounds like our GCC won't work with mold even after we ship this package, which isn't good. We should really look into the GCC issue!

PR #2365, opened 2023-03-21
addpatch: bin86

r-value:
Add definition of `Long` as `int` for riscv64.


r-value:
The problem looks distro-specific because the source code is from Arch Linux and differs from upstream code repo. It might be difficult to upstream the patch.

r-value:
Upstreamed here: jbruchon/dev86#25

felixonmars:
Thanks!

PR #2367, opened 2023-03-21
addpatch: blop

r-value:
Update config.guess


felixonmars:
Please also try to contact upstream so we don't need this in the future :)

r-value:
> Please also try to contact upstream so we don't need this in the future :)

It seems there is no contact info available for upstream and it's never updated since 2007 :(

felixonmars:
> > Please also try to contact upstream so we don't need this in the future :)
> 
> It seems there is no contact info available for upstream and it's never updated since 2007 :(

https://sourceforge.net/p/blop/bugs/ may worth a note :)

PR #2372, opened 2023-03-23
upgpatch: stylelint

kxxt:
Fix rotten patch.


kxxt:
Patch updated. Please merge this after #2463

PR #2388, opened 2023-03-28
addpatch: m17n-db

sihuan:


felixonmars:
Please also try to contact upstream so we don't need this in the future :)

sihuan:
[Reported to upstream via Email.](https://lists.nongnu.org/archive/html/m17n-list/2023-03/msg00003.html)

PR #2392, opened 2023-03-29
upgpatch: python-eventlet

hleft:
build get `AttributeError: module 'dns.rdtypes' has no attribute 'ANY'`

fix in eventlet  v0.33.3 release


Xeonacid:
We shouldn't bump to version newer than Arch and wait for Arch's upgrade.

hleft:
@Xeonacid  ok, now still use patch

python-eventlet/riscv64.patch
```diff
@@ -1,29 +1,31 @@
...
+@@ -20,17 +20,25 @@ optdepends=(
  )
  options=('!makeflags')
  source=(https://github.com/eventlet/eventlet/archive/v${pkgver}/${_pyname}-${pkgver}.tar.gz
 -        python310.patch)
 +        python310.patch
-+        "increase-test-thread-sleep-time.patch")
- sha512sums=('437deffb2e24a3361585a5861d35bc19131d80ffd54fb8172e232683286fc0f28e68fd1a89f7ebe5067a5b9699b6c3382c9be5868adc2ebb11c2d205b644093c'
++        "increase-test-thread-sleep-time.patch"
++        "greendns.patch")
```
Xeonacid, line 10: Should use remote url when available instead of storing local patch for better tracking.

PR #2400, opened 2023-03-31
addpkg: dsq

Ast-x64:


XieJiSS:
Please consider also reporting this to upstream :-)

xctan:
I found another similar issue https://github.com/multiprocessio/dsq/issues/80, where the author suggests that this test case can be safely disabled.

Xeonacid:
Superseded by #3168

PR #2449, opened 2023-04-03
updpatch: gcc

r-value:
`gcc` delibrately filters out default library paths when invoking linkers, causing some linkers like `mold` fails to find libraries.
This patch is a workaround and will be proposed for upstream soon.

See-also:
rui314/mold#358
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104707


CoelacanthusHex:
There is no need to keep saving the space of the parameters, since all modern operating systems have a large enough parameter space （all of them has 262k at least).

PR #2459, opened 2023-04-04
addpatch: webkit2gtk

hack3ric:
Backported riscv64 build fix from master, with modification

Source: https://commits.webkit.org/261498@main


hack3ric:
Arch upstream on why they didn't merge webkitgtk packages into one split package: https://bugs.archlinux.org/task/78110

PR #2487, opened 2023-04-08
addpatch: cjdns

hack3ric:
- Replace riscv64gc with riscv64 in TARGET and HOST (that passes into C compilation) in `libsodium-sys`
- Enable riscv64 build in `cjdns_sys`


hack3ric:
Upstreamed: https://github.com/cjdelisle/cjdns/pull/1250

felixonmars:
Actually libsodium-sys supports building with system libsodium instead (specifying SODIUM_USE_PKG_CONFIG), and should be preferred on Arch. This could be raised to Arch upstream too :)

hack3ric:
The C part of this package actually links to the vendored libsodium from Rust, instead of trying to find one using pkgconf. It is hard to decide a fix since there's no good way of detecting Rust features from C, but I could propose a fix at a later date.

Issue #2501, opened 2023-04-12
Add CI test support for bazaar source


a1ca7raz:
I successfully fixed [libappindicator](https://github.com/felixonmars/archriscv-packages/pull/2500). This package fetches source code from launchpad via breezy.
But it [failed](https://github.com/felixonmars/archriscv-packages/actions/runs/4670769644/jobs/8270910927) to pass the CI test. The test does not support breezy and fetching from launchpad via breezy.
I have fixed the test script and CI workflow. Is it possible to merge the pr?

a1ca7raz:
I reported the missing dep to archlinux. libappindicator should be ok soon.
https://bugs.archlinux.org/task/78176
But breezy is still missing in make dependencies of CI test.

Xeonacid:
> But breezy is still missing in make dependencies of CI test.

You may submit a PR to install `devtools` as mentioned before.

PR #2502, opened 2023-04-12
fixci: add bazaar support for CI test

a1ca7raz:
See https://github.com/felixonmars/archriscv-packages/issues/2501


CoelacanthusHex:
According to discussion in internal group, prefer to install `devtools` package to satisfy all depends we need.

PR #2510, opened 2023-04-14
addpatch: deno

hack3ric:
- Patch rusty_v8 to allow passing custom GN arguments, as well as disabling printing GN arguments (it segfaults, but it is trivial for now since it only affects printing arguments, not generating) [1]

- Use system clang, lld, gn and ninja to build V8 from scratch, since there's no riscv64 prebuilt libraries available.

- Enable `v8_enable_shared_ro_heap`. This is intentially disabled by upstream [2], but we need to wait for pointer compression to meet rusty_v8's requirement [3] and disable it

[1]: https://github.com/denoland/rusty_v8/pull/1209
[2]: https://github.com/denoland/rusty_v8/blob/5dce1eaeef6457b7571b12c9c6170a362a62bb38/.gn#L52
[3]: https://github.com/denoland/rusty_v8/blob/5dce1eaeef6457b7571b12c9c6170a362a62bb38/src/binding.cc#L2907

### DIFF
--- a/deno/riscv64.patch
+++ b/deno/riscv64.patch
@@ -1,0 +1,49 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -11,17 +11,42 @@ url="https://deno.land"
+ license=('MIT')
+ options=('!lto')
+ depends=('gcc-libs')
+-makedepends=('git' 'python' 'cargo' 'nodejs')
+-source=("git+https://github.com/denoland/deno.git#commit=$_commit")
+-sha512sums=('SKIP')
++makedepends=('git' 'python' 'cargo' 'nodejs' 'gn' 'ninja' 'clang' 'lld')
++source=("git+https://github.com/denoland/deno.git#commit=$_commit"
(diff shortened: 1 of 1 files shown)


hack3ric:
Note that current build runs TypeScript scripts normally, but in REPL errors will appear (while not affecting basic auto completion functionality):

```
ERROR TSLS = Error: Could not find source file: 'file://<working_dir>/$deno$repl.ts'.
    at getValidSourceFile (ext:deno_tsc/00_typescript.js:133358:23)
    at Object.getSemanticDiagnostics (ext:deno_tsc/00_typescript.js:133600:32)
    at serverRequest (ext:deno_tsc/99_main_compiler.js:1059:34)
    at [ext:cli/lsp/tsc.rs:3496:26]:1:12
```

This is triggered only by TypeScript LSP's `getDiagnostics` request. Other LSP requests with specifiers containing `$deno$repl.ts` seems to not fail. I'm yet to find out why.

PR #2525, opened 2023-04-15
addpkg: couchdb

TinySnow:
Fixed `package()` error.

The origin PKGBUILD file hardcoded the version number. While building, system can't find the files.

I can't find some bash variables can replace the version number, ($pkgver is not corresponding) so I just fix it to right one. However it's also hardcoding.

I don't know whether this dispose is appropriate or not, if exists better one, please teach, thanks.

Build passed.

### DIFF
--- a/couchdb/riscv64.patch
+++ b/couchdb/riscv64.patch
@@ -1,0 +1,11 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -75,7 +75,7 @@ package() {
+   install -Dm644 ${pkgname}.sysusers "${pkgdir}"/usr/lib/sysusers.d/${pkgname}.conf
+ 
+   # Remove some cruft
+-  rm -r "${pkgdir}"/usr/lib/couchdb/erts-13.1.1/{doc,include,lib,man,src}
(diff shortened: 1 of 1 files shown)


Xeonacid:
Is it also a bug on x86_64? If so, please report to Arch. If not, please examine why.

TinySnow:
> Is it also a bug on x86_64? If so, please report to Arch. If not, please examine why.

Yes, it is. I test it by trying to port via `extra-x86_64-build`. Error occurred at the same place that the log displayed.

Tomorrow to report on Arch.

BTW, is it right to hardcode here?

XieJiSS:
Since the original PKGBUILD hard-coded the version string, keep the old way won't make things worse. That said, I believe there are some elegant ways to get the version substring at runtime. You can have a try in case your time and ability are capable (but that's not a strict requirement).

TinySnow:
OK, got it.

After reporting to Arch, the link will be post here to keep consistency.

TinySnow:
Task on Arch Linux Bug Tracker: https://bugs.archlinux.org/task/78215.

PR #2529, opened 2023-04-15
addpkg: xosd

Suyun114:
Fixed `config.guess`.

The [upstream](https://sourceforge.net/projects/libxosd/) does not have a ticket tracker to create a bug report.


Xeonacid:
You may ask Arch to change source to https://github.com/AndreRenaud/XOSD, which is owned by its main author and has a newer version.

Suyun114:
> You may ask Arch to change source to https://github.com/AndreRenaud/XOSD, which is owned by its main author and has a newer version.

https://bugs.archlinux.org/task/78211

PR #2542, opened 2023-04-18
addpkg: dpf-plugins

A1ca7raz:
Add missing make dependency.


a1ca7raz:
> It seems that this is also an issue for x86_64.

Reported to arch.
https://bugs.archlinux.org/task/78252

PR #2549, opened 2023-04-20
upgpatch: gifski

hack3ric:
GitHub seems to have modified their commit patch, so vendor it instead.


Xeonacid:
You may use https://github.com/zmwangx/rust-ffmpeg/commit/d5e9f34b6a3cf309178389ac996b6ff0130a2cb7.diff instead of `.patch` to avoid checksum change.

hack3ric:
Done.

PR #2574, opened 2023-04-29
devtools-riscv64: use https://riscv.mirror.pkgbuild.com as default mirror

CoelacanthusHex:
Get better performance everywhere. And the mirror synchronization frequency is high enough.

As we do in rootfs.

### DIFF
--- a/devtools-riscv64/pacman-extra-riscv64.patch
+++ b/devtools-riscv64/pacman-extra-riscv64.patch
@@ -14,23 +14,23 @@
  
  #[testing]
 -#Include = /etc/pacman.d/mirrorlist
-+#Server = https://archriscv.felixc.at/repo/$repo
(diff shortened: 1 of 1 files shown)


felixonmars:
This would unfortunately break rebuilds though. Until we have a way to override the mirrors per builder, the default mirror is still unfortunately the only choice.

CoelacanthusHex:
Ok, I will close it and open an issue to track it.

XieJiSS:
I'm wondering why this would break rebuilds, is it related to the lag between the default site and mirrors?

felixonmars:
> I'm wondering why this would break rebuilds, is it related to the lag between the default site and mirrors?

Yes. Builders using a not up-to-date may rebuild against an old version of a just-updated dependency.

PR #2578, opened 2023-04-29
addpkg: koko

Integral-Tech:


koko/riscv64.patch
```diff
@@ -0,0 +1,24 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -5,7 +5,7 @@ pkgver=23.04.0
+ pkgrel=1
+ pkgdesc='Image gallery application'
+ url='https://apps.kde.org/koko/'
+-arch=(x86_64 riscv64)
++arch=(x86_64)
```
Xeonacid, line 8: Is this committed by mistake? (We don't need the reversing `arch` change also here)

Xeonacid (changes requested):
Please report the checksum mismatch to Arch.

koko/riscv64.patch
Integral-Tech: Yes, it is committed by mistake :(

Integral-Tech:
> Please report the checksum mismatch to Arch.

Therefore, should I close this pull request?

Xeonacid:
Sorry for my misunderstanding words. We could merge it here for a quick fix, but should **also** report it to Arch. :)

Integral-Tech:
> Sorry for my misunderstanding words. We could merge it here for a quick fix, but should **also** report it to Arch. :)

OK :)

Xeonacid (changes requested):
Check still fails:
```text
patch: **** malformed patch at line 10: @@ -16,9 +16,9 @@ source=(https://download.kde.org/stable/release-service/$pkgver/src/$pkgname-$pk
```

r-value:
Please re-generate the patch instead of modifying it manually.

Xeonacid (approved):

felixonmars:
Upstream report at https://bugs.archlinux.org/task/78432

PR #2579, opened 2023-04-29
upgpatch: mysql-workbench

kxxt:
Fix rotten patch.
The build error that is fixed by #1270 does not occur in this version.


XieJiSS (approved):
lgtm

PR #2583, opened 2023-05-01
addpkg: java8-openjdk-bootstrap

kxxt:
Cross compile openjdk8 (with zero jvm) from x86_64.

The java8-openjdk-add-riscv-support.patch is taken from https://gitee.com/misaka00251/openjdk-1.8.0/blob/master/add-riscv-support.patch

We may try to enable the JIT later, using this patch: https://gitee.com/misaka00251/openjdk-1.8.0/blob/nobisheng/riscv64.patch , or this repo: https://github.com/zhangxiang-plct/jdk8u

After cross compiling the bootstrap packages, java8-openjdk can be bootstrapped on riscv64.

I tested the final packages by building and testing `java-commons-lang` package. Only one test failed with timeout. Previously building this package with jdk19 failed with 96 failures.
(https://archriscv.felixc.at/.status/log.htm?url=logs/java-commons-lang/java-commons-lang-3.12.0-1.log). So this may suggest the riscv jdk19 is buggy.

### DIFF
--- a/java8-openjdk-bootstrap/PKGBUILD
+++ b/java8-openjdk-bootstrap/PKGBUILD
@@ -1,0 +1,341 @@
+# Maintainer: Levente Polyak <anthraxx[at]archlinux[dot]org>
+# Maintainer: Frederik Schwan <freswa at archlinux dot org>
+# Maintainer: Guillaume ALAUX <guillaume@archlinux.org>
+# Contributor: Boyan Ding <stu_dby@126.com>
+
+pkgname=('jre8-openjdk-bootstrap-headless' 'jre8-openjdk-bootstrap' 'jdk8-openjdk-bootstrap')
+pkgbase=java8-openjdk-bootstrap
+
+_majorver=8
+_minorver=372
+_updatever=07
+pkgver=${_majorver}.${_minorver}.u${_updatever}
+pkgrel=1
+arch=(any)
+url='https://openjdk.java.net/'
+license=('custom')
+makedepends=(
+  'bash'
+  'ccache'
+  'cpio'
+  'git'
+  'java-environment=8'
+  'unzip'
+  'zip'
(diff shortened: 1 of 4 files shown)


XieJiSS (approved):
Nice catch on the flaky jdk19! We may need to investigate it later. Also would it be possible to use upstream patches directly? Like putting their urls directly inside the source array so that devtools will download them automatically. This way we don't need to maintain a copy in our repo.

Xeonacid (commented):
It's much better to `pacstrap` a new rootfs, instead of mannually downloading and extracting packages.

kxxt:
@XieJiSS Thanks for the approval.

> Also would it be possible to use upstream patches directly? Like putting their urls inside the source array so that devtools will download them automatically.

I don't know if putting a gitee url in the source array is appropriate so I added the patch to this PR in the first place.

Several issues regarding referencing a patch in gitee repo:
- Downloading big file needs authentication, not a case for `add-riscv-support.patch`, but for [`riscv64.patch`](https://gitee.com/misaka00251/openjdk-1.8.0/blob/nobisheng/riscv64.patch) that might be used in the future.
- It is a fork of https://gitee.com/src-openeuler/openjdk-1.8.0, I am not sure if the fork repo will always exist in the future.

Tell me if you think referencing the patch via this url is appropriate:
https://gitee.com/misaka00251/openjdk-1.8.0/raw/e4b9d135939e2263146905911ad100424101e305/add-riscv-support.patch

CoelacanthusHex:
> It's much better to `pacstrap` a new rootfs, instead of mannually downloading and extracting packages.

We just need correct directory structure to make sure jdk and libraries in sysroot work, not necessary to use pacstrap so we needn't deal with root permission problem.

kxxt:
Now all the patches are not stored in this repo.

PR #2584, opened 2023-05-01
addpatch: java8-openjdk

kxxt:
Add RISC-V support(use zero jvm).


XieJiSS (commented):
Similar to the previous PR - would it be possible to specify a link pointing to the patch inside source array instead of submitting a copy of the patch? Thanks!

kxxt:
Now the patch is not stored in the repo.

PR #2597, opened 2023-05-08
updpkg: meson 1.1.0

r-value:
* REQUIRES SKIPPING CHECKS because `ldc` is not ready yet
 * Some other tests also fails because of outdated `gobject-introspection`, but this package depends on `meson`, forming a circular dependency. So just skip the test for now.


XieJiSS (approved):
Confirmed that the list of removed packages is identical to the previous version

PR #2598, opened 2023-05-09
upgpatch: python-nodeenv

hack3ric:
- Use .diff rather than .patch in GitHub PR patch for consistent content
- Use system nodejs since riscv64 prebuilt of latest version may not be present


XieJiSS (approved):
Should we also submit this change (namely the "use system node" patch) to Arch upstream?

felixonmars:
I think so. It's a pity that using system node.js requires a patch though. Would be great if this is doable via something like an environment variable.

PR #2602, opened 2023-05-10
qemu-user-blacklist: add digikam

hack3ric:
The package checks SSE availability by reading /proc/info, which is inherited from host machine in nspawn.


hack3ric:
/proc/cpuinfo*

PR #2629, opened 2023-05-12
updpatch: rust 1.69.0

r-value:
* Bump dependency `cc-rs` of `bootstrap` for musl-gcc to work correctly.


FantasqueX:
Should we upstream this version bump?

r-value:
> Should we upstream this version bump?

Notice that this bump is not tested sufficiently against other architectures yet.
There's no hurry. I'm working on this. It won't be released earlier than July even if it has been merged now.

FantasqueX:
> > Should we upstream this version bump?
> 
> Notice that this bump is not tested sufficiently against other architectures yet. 

So, is there any links to upstream to discuss this version bump?

> There's no hurry. I'm working on this. It won't be released earlier than July even if it has been merged now.

Does it mean that this PR is still a draft?

r-value:
> > There's no hurry. I'm working on this. It won't be released earlier than July even if it has been merged now.
> 
> Does it mean that this PR is still a draft?

I can't get your point here. First, it's working in RISC-V doesn't mean it will work in other arch. Second, this PR is a proposal for *archriscv*.

FantasqueX:
Still want to leave an upstreamed link https://github.com/rust-lang/rust/pull/111701

PR #2671, opened 2023-05-30
rmvpatch: zabbix

iamtwz:
`uname_uint8.go` does not exist in upstream.

The implementation method has been changed by the upstream, and it can be compiled successfully.


felixonmars:
It seems the removed patch is for `zabbix` instead of `ming`.

iamtwz:
> It seems the removed patch is for `zabbix` instead of `ming`.

Sorry for typo, I've already changed the commit message and pull request title.

PR #2696, opened 2023-06-06
addpatch: arduino

kxxt:
- Patch build.xml to add riscv64 support
- No prebuilt astyle and listSerialPortsC available so we build them ourselves.
  - Pin the source to a commit because upstream didn't create new tags for them.
- I prefer not to upstream the patch as upstream previously didn't accept the powerpc port.


Xeonacid:
> * I prefer not to upstream the patch as upstream previously didn't accept the powerpc port.

It is still worth trying to upstream.

PR #2720, opened 2023-06-10
addpkg: startup-notification

TinySnow:
Fixed `config.guess`.

upstream url: https://gitlab.freedesktop.org/xdg/startup-notification/-/issues/5.


Xeonacid (changes requested):
The patch check is failing...

TinySnow:
> The patch check is failing...

`asp` pulled the old version, that's the reason why check failed. I'll modify the patch, at the same time, apply to fix the half-automatic script in `wiki` to `pkgctl` in some days. Please give me some time.

Xeonacid (approved):

TinySnow:
> The patch check is failing...

The new patch test passed.

PR #2754, opened 2023-06-24
addpkg: festival

A1ca7raz:
Fix config.guess&config.sub


felixonmars:
Please try to upstream the change so we don't have to keep it downstream forever :)

PR #2757, opened 2023-06-24
addpkg: open-isns

A1ca7raz:
Fix config.guess&config.sub


felixonmars:
Please try to upstream the change so we don't have to keep it downstream forever :)

felixonmars:
Since 0.102 upstream switches to meson and this shouldn't be a problem anymore in the future. Opened https://bugs.archlinux.org/task/79539 to ask Arch upstream for changing.

Let's merge it for now and we could remove the patch later when Arch switched.

PR #2765, opened 2023-06-26
addpatch: gcc12

hack3ric:
Reused previous gcc patch with modifications.


hack3ric:
The patch is ready now.

PR #2774, opened 2023-06-29
addpatch: electron22

kxxt:
Prerequisite: rebuild python-httplib2 against python 3.11 

- Apply patches from OpenSUSE
    - riscv patches
    - swiftshader patches
        - Use system llvm because the bundled llvm 10 won't work.
        - There are some llvm16 related patches in the SUSE repo. We don't need them for now.
- Disable pulling of prebuilt binaries.
    - Use packaged versions if possible. 
    - Build some of them ourselves.
- Replace the bundled & hardcoded x86_64 JDK11 with our system JDK11.


kxxt:
PR updated. It now uses a python script to determine the commit hash of luci_go repo.

kxxt:
Update: fix typos.

PR #2776, opened 2023-06-30
addpkg: obs-studio

r-value:
`char` is unsigned on RISC-V and it's impl-defined for `char` to be signed or unsigned. Ensuring `utf8_char_t` to be signed to make comparison with zero works.


Xeonacid (commented):
Patch test is failing.

r-value:
Fixed and follows upstream PR https://github.com/obsproject/obs-studio/pull/9184

PR #2791, opened 2023-07-06
addpatch: code

kxxt:
- Add missing riscv64 definition.
- Skip browser downloads(they are not used at all).
- Skip electron binary download. (No upstream binary available)
- Create a zipped electron22 and put it in the cache to make the build process happy.
- Build all native node extensions in debug mode to avoid segfaults and strange behaviors.
    - This bug needs further investigation.


kxxt:
It is recommended to run code with `--no-sandbox`  for now to avoid bugs.

BTW the electron22 in our repo is not working, which will be fixed in an upcoming PR.

PR #2806, opened 2023-07-13
upgpatch: electron22

kxxt:
This PR fixes electron22 to make Code OSS work on riscv64 (still needs `--no-sandbox`).

The full PKGBUILD is [here](https://bin.garudalinux.org/?99ed90d41c76b4b9#8N7gaBo7yVv5jUxkdgi6Ag9esyUip31XRia9FHMkgDmb) for reference.

# Changes

- Add an option to select build profile. `Testing` profile is for debug purpose. It is much eaiser to debug.
- Lock esbuild version instead of using system-wide one because the build process checks its version.
- Add two patches that fix build errors occurred in `Testing` profile.
- Replace old crashpad patch with chromium upstream one. (The old patch failed many checks in `Testing` profile.)
- Add another two patches that fill in missing riscv64 cases.
- Don't use compiler flags and linker flags from `/etc/makepkg.conf`. They will cause electron to crash with weird errors.
- Set symbol_level to 1 for symbols in the backtrace. (See https://www.chromium.org/developers/gn-build-configuration/#faster-builds-with-no-or-minimal-symbols)

# SG2042 Workarounds

Many programs randomly hangs on sg2042. If it hangs, just interrupt the build and rerun it.

Node.js 16 randomly segfaults on sg2042 (and might bring sg2042 down sometimes). The following workaround builds the targets that needs node.js first and retries it until success.

```bash
local _prebuild_targets=$(ninja -C out/$_build_profile -t targets all | grep -E "^gen/third_party/devtools-frontend/" | sed -e 's/:.*//g')  
until printf '%s\n' "$_prebuild_targets" | xargs -d '\n' ninja -C out/$_build_profile  
do  
    echo "Failed to build devtools-frontend. Retrying..."  
    sleep 1  
done
```

# Debugging Electron

- Use testing build by setting `_build_profile` to `Testing`.
-  (Optional) Enable all checks by setting GN arg `dcheck_always_on` to `true`.
- Incremental build instead of clean build.

# ToDos for the Future

- Investigate what compiler flags in `/etc/makepkg.conf` is causing built electron to crash with weird errors.
- Opening the `Files` menu will hit [a check in `Testing` build](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:ui/base/accelerators/accelerator.cc;drc=6754d1409bf5099314eea7e87e896622ade9bc0f;l=467). (Only for the `Files` menu, other menus won't hit this) This needs further investigation. In release build, it seems that the check is optimized away.
- Several parts of chromium is missing riscv64 implementations, though they are not critical. e.g. user_agent.cc
- Segfault happens when running several electron22 test cases (with sandbox disabled). It needs more debugging.
- Investigate failing tests.


kxxt:
Test logs are available at https://bin.garudalinux.org/?6aa76271955fc226#DUEdA4djE4Aqak3Kfzruo5gc8nyVSniBkC7B6vTL19rG

Test summary:

tests 2109
pass 2076
fail 33

PR #2864, opened 2023-07-30
addpatch: jenkins

hack3ric:
Use exec-maven-plugin to execute system Yarn.


Xeonacid:
This might be reported to Arch as well.

hack3ric:
> This might be reported to Arch as well.

Node and Yarn are only used at build time. Probably Arch wouldn't bother maintaining a patch when the upstream code works well on x86.

PR #2866, opened 2023-07-30
addpatch: teeworlds

hack3ric:
Allow build on `unknown` architecture.


felixonmars:
We probably should make it detect riscv64 correctly instead.

hack3ric:
> We probably should make it detect riscv64 correctly instead.

The platform detection is done in `bam`. I'll create a new PR for it and change `unknown` to `riscv64` in this one.

PR #2871, opened 2023-07-31
addpatch: bam

hack3ric:
Add platform definition for riscv64.


felixonmars:
Submitted as https://github.com/matricks/bam/pull/158

PR #2879, opened 2023-08-08
updpatch: rust 1:1.71.1-1

hack3ric:
Fix rotten patch.


rust/riscv64.patch
```diff
@@ -103,26 +101,3 @@ index 99f0b93..56bbbc7 100644
...
--  provides=(lib32-rust)
--  conflicts=(lib32-rust)
--  replaces=(lib32-rust)
--
--  cp -a dest-i686/* "$pkgdir"
--
--  mkdir -p "$pkgdir/usr/share/licenses"
--  ln -s rust "$pkgdir/usr/share/licenses/$pkgname"
--}
--
```
r-value, line 125: why are these removals removed?

hack3ric: The `package_$name` functions never gets called if `$pkgname` does not contain `$name`.

r-value: I think we'd better keep the `PKGBUILD` clean

hack3ric: Added back.

Xeonacid: > I think we'd better keep the `PKGBUILD` clean

I think it would be better to keep a minimal patch for less rotten patch.

PR #2880, opened 2023-08-08
addpatch: pwru

hack3ric:
Use generated vmlinux.h on riscv64 and enables build on riscv64.

Will upstream with more changes unrelated to current build.


hack3ric:
This still has problems - It requires a certain set of kernel parameters to be set in order to work, and it seems to freeze on futex_wait on Unmatched.

The ebpf patch is upstreamed in https://github.com/cilium/ebpf/pull/1110, but more testing is needed for the main patch to be upstreamed.

PR #2922, opened 2023-09-01
updpatch: electron23

hack3ric:
Add missing swiftshader patch.


electron23/electron23-swiftshader-llvm-15.patch
kxxt, line 1: Arch has upgraded llvm to 16, is this really needed?

PR #2926, opened 2023-09-02
updpatch: deno 1.36.3-1

hack3ric:
Revert use of zlib-ng upstream [1] due to RVV detection problems.

[1]: https://github.com/denoland/deno/commit/85a2b281f566d3404d23852ae29d4a75d020dd5e


felixonmars:
Do we have a way to disable RVV in zlib-ng instead?

PR #2949, opened 2023-09-07
addpatch: opensubdiv 3.5.0-2

Xeonacid:
- Disable CUDA
- Fix patch checksum mismatch (no need to report to Arch I think)


felixonmars:
> * no need to report to Arch I think

Actually I disagree here. It's always an Arch bug if the PKGBUILD doesn't work on x86_64 :)

PR #2951, opened 2023-09-10
addpatch: retroarch

sajattack:
Enables some options to support more weak graphics apis and disables ffmpeg and gamemode, which were causing issues for me.

### DIFF
--- a/retroarch/riscv64.patch
+++ b/retroarch/riscv64.patch
@@ -1,0 +1,38 @@
+--- PKGBUILD	2023-09-10 00:55:26.455719410 -0700
++++ PKGBUILD-riscv	2023-09-10 00:52:21.902014518 -0700
+@@ -6,13 +6,12 @@
+ pkgver=1.15.0
+ pkgrel=3
+ pkgdesc='Reference frontend for the libretro API'
+-arch=(x86_64)
++arch=('x86_64' 'riscv64')
+ url=http://www.libretro.com/
+ license=(GPL)
+ groups=(libretro)
+ depends=(
+   alsa-lib
+-  ffmpeg
+   flac
+   libass.so
+   libdrm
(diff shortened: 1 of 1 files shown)


Xeonacid:
> and disables ffmpeg and gamemode, which were causing issues for me.

Could you explain what issue you are facing for a backlog? Thanks!

retroarch/riscv64.patch
```diff
@@ -0,0 +1,38 @@
+--- PKGBUILD	2023-09-10 00:55:26.455719410 -0700
++++ PKGBUILD-riscv	2023-09-10 00:52:21.902014518 -0700
+@@ -6,13 +6,12 @@
+ pkgver=1.15.0
+ pkgrel=3
+ pkgdesc='Reference frontend for the libretro API'
+-arch=(x86_64)
++arch=('x86_64' 'riscv64')
```
Xeonacid, line 8: We don't need to change here for a minimal patch, our build script will do this automatically. :)

sajattack:
> > and disables ffmpeg and gamemode, which were causing issues for me.
> 
> Could you explain what issue you are facing for a backlog? Thanks!

ffmpeg was complaining about opencl linkage which I might have broken unrelatedly. It might not be necessary for this patch to exclude ffmpeg. However it seems like a bulky dependency that probably adds little value. A package for gamemode doesn't exist.

Xeonacid:
> A package for gamemode doesn't exist.

Do you mean the [`gamemode` package](https://archriscv.felixc.at/repo/extra/gamemode-1.7-1-riscv64.pkg.tar.zst)? It exists in our repository though I don't test if it works fine. 🤔

sajattack:
Ok I'll spend some more time investigating gamemode and ffmpeg another time. Maybe next weekend.

PR #2975, opened 2023-09-21
rmvpatch: mkosi

hack3ric:
Upstream released new versions with the fix.


hack3ric:
`package()` uses pandoc which is outdated due to recent Haskell update. We should merge this after pandoc has updated.

PR #3004, opened 2023-09-23
updpatch: fwupd 1.9.5-2

hack3ric:
Fix rotten patch.

### DIFF
--- a/fwupd/riscv64.patch
+++ b/fwupd/riscv64.patch
@@ -1,32 +1,29 @@
-diff --git PKGBUILD PKGBUILD
-index c69d56f..0deea56 100644
 --- PKGBUILD
 +++ PKGBUILD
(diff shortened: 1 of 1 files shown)


fwupd/riscv64.patch
```diff
@@ -1,32 +1,30 @@
...
+   libxmlb
+   polkit
+   python
+@@ -81,7 +80,10 @@ build() {
+   arch-meson ${pkgname}-${pkgver} build \
+     -D b_lto=false \
+     -D docs=enabled \
+-    -D plugin_intel_spi=true \
++    -D plugin_intel_spi=false \
++    -D plugin_intel_spi=false \
```
felixonmars, line 17: plugin_intel_spi=false appears to be duplicated after the change.

felixonmars (changes requested):

fwupd/riscv64.patch
hack3ric: Fixed.

PR #3020, opened 2023-09-24
addpkg: yodl

moui0:
I found a strange thing: building yodl requires icmake, but under the riscv64, icmake requires loading `libbobcat.so.5`, which doesn't seem to happen on x86_64.

This is the error message:

```
==> Starting build()...
/usr/bin/icmake: error while loading shared libraries: libbobcat.so.5: cannot open shared object file: No such file or directory
==> ERROR: A failure occurred in build().
```

I took an ugly approach to solve this problem: Created `libbobcat.so.5` from `libbobcat.so.6` through a symbolic link. (`libbobcat.so.6` is a dependency of icmake).

If you have a better way to fix it, please let me know.


CoelacanthusHex:
It should be icmake need rebuild.

r-value (changes requested):
This loader error is caused by outdated `icmake` binaries. Please wait for it to be rebuilt against libbobcat 6 before trying to build new package with it.

XieJiSS:
(That said, this PR still counts as one point)

felixonmars:
Superseded by https://github.com/felixonmars/archriscv-packages/pull/3026 with proper bootstrapping.

Thanks for working on this.

PR #3029, opened 2023-09-25
addpkg: maude

moui0:
Update the outdated config.guess file.

Perhaps there is no need to report to upstream, because maude is going to use CMake instread of Autotools: https://github.com/SRI-CSL/Maude/pull/10 ?

### DIFF
--- a/maude/riscv64.patch
+++ b/maude/riscv64.patch
@@ -1,0 +1,11 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -25,7 +25,7 @@ prepare() {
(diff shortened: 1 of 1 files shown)


Xeonacid:
> Perhaps there is no need to report to upstream, because maude is going to use CMake instread of Autotools: [SRI-CSL/Maude#10](https://github.com/SRI-CSL/Maude/pull/10) ?

It seems a community-made PR, which is huge and WIP. Whether upstream would accept it is not clear. So I think it is still worth reporting.

PR #3042, opened 2023-09-29
upgpatch: i3status-rust 0.32.2-1

moui0:
Fix rotten patch.


felixonmars:
I don't think this is necessary. The current patch still works.

PR #3043, opened 2023-09-29
addpkg: python-mitmproxy-rs

moui0:
Fix `ring` issue and add RISC-V 64 support for `protoc-bin-vendored` (Upstreamed to
https://github.com/stepancheg/rust-protoc-bin-vendored/pull/2)


felixonmars:
While this is an excellent fix, I would still prefer if we could utilize system protobuf package instead of downloading a binary. This could also be raised to Arch Linux if the change is simple enough (like for example, adding protobuf as makedep and probably setting an environment variable).

moui0:
Got it, thanks for your patience.

Reported to Arch Linux: https://bugs.archlinux.org/task/79823

PR #3070, opened 2023-10-09
addpkg: p2pool

LithiumTins:
The author use plenty of casting like `(uint64_t*)arr` in which `arr` is a array of `uint8_t*`. Meanwhile, the author use compile option `Wcast-align` and `Werror`. I found it interesting that the warning isn't caused on x86_64 but it is on riscv64 so the build failed. So I use #pragma to block the warning temporarily from files which contain the type of cast. 

And there's another problem caused by cast-qual in the external project RandomX, which happens when cast from const to non-const. However it seems not necessary to remove the const qualifier, so I opened a issue: https://github.com/tevador/RandomX/issues/277 and the problem is fixed on the master branch. Maybe it will take effect in a future release.

### DIFF
--- a/p2pool/riscv64.patch
+++ b/p2pool/riscv64.patch
@@ -1,0 +1,31 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -10,8 +10,12 @@ license=('GPL-3.0-or-later')
+ url="https://github.com/SChernykh/p2pool"
+ depends=('curl' 'libuv' 'libsodium' 'zeromq')
+ makedepends=('cmake' 'git')
+-source=("${pkgname}"::"git+https://github.com/SChernykh/${pkgname}#commit=${_commit}")
+-b2sums=('SKIP')
++source=("${pkgname}"::"git+https://github.com/SChernykh/${pkgname}#commit=${_commit}"
++        "https://github.com/SChernykh/p2pool/commit/5daa2b9dfffcef72286b8429d4c3d65f535a70a8.patch"
++        "https://github.com/SChernykh/RandomX/commit/d5035cd9c50a5d34ae87ec3507314dc806343548.patch")
++b2sums=('SKIP'
++        'c5aa360cf5532476e415c2b30517c2bc2283b615d88c3dffcee5f06b303c0801c0e31f3e40959b167cc8d26c2df76c0cfbddf617c3571a04d2b5c3dcaca17d35'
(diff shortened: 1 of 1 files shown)


r-value:
Before any discussions about the fix itself, please don't use `sed` to modify sources which is almost unreadable. We prefer a patch file for source code modifications at this scale.

r-value:
Since it's already fixed in commit https://github.com/SChernykh/RandomX/commit/7167cf1ce6e5e9b240967e0d96148966a24bbdb0, we'd better cherry-pick that commit as a patch as a temporary solution and remove it later when it's released.

r-value:
For the first problem, as described in the [gcc documentation]:

> `-Wcast-align`
> Warn whenever a pointer is cast such that the required alignment of the target is increased. For example, warn if a char * is cast to an int * on machines where integers can only be accessed at two- or four-byte boundaries.

So it's arch-dependent by design. x86-64 doesn't care much about misaligned memory accesses so it never generates the warning, while on RISC-V it does. To port the code correctly, you'll need to modify the code so that it will be aligned perfectly in any situation before suppressing the warning because it does make sense on RISC-V.

XieJiSS:
As a general conclusion, this attempt is a good first step for solving this issue, and we would like to encourage you to keep working on this PR to 1) cherry-pick the upstream commit as a patch file, and apply the patch in PKGBUILD (please refer to previous PRs), and 2) read & investigate the upstream fix, and do some searchings about misaligned mem access behavior to gain some insights about useful background knowledge, especially when it comes to cross-architectural difference.

r-value:
After some investigations by myself, I'll recommend to open an issue describing the unnecessary `-Wcast-aligned` because upstream has already implemented [`read_unaligned`](https://github.com/SChernykh/p2pool/blob/011d05f7b68df4bd215e7f9aa459cc6e777cb063/src/util.h#L155) function in such a way that the warning will be an unavoidable false positive.

LithiumTins:
Thanks for your detailed recommendation. I'll continue working on this pr and try to follow the advice above.

XieJiSS:
Looks great! Now the only thing we'd need to do is to squash these two commits into one, so that the commit history remains clean. Also, as a common practice, we do not provide details about commit in its commit message but put them in PR description instead. You can edit your commit messages when you are rebasing / squashing them.

r-value (approved):
LGTM, thank you for your contribution!

PR #3127, opened 2023-10-18
addpatch: virt-viewer

aimixsaka:
fix source url
[upstream bug report](https://bugs.archlinux.org/task/79994)

### DIFF
--- a/virt-viewer/riscv64.patch
+++ b/virt-viewer/riscv64.patch
@@ -1,0 +1,16 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -13,7 +13,7 @@ license=('GPL')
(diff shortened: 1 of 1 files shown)


Xeonacid:
Patch CI fails. We should checkout [last tag](https://gitlab.archlinux.org/archlinux/packaging/packages/virt-viewer/-/commits/11.0-2?ref_type=tags) instead of HEAD.

aimixsaka:
> Patch CI fails. We should checkout [last tag](https://gitlab.archlinux.org/archlinux/packaging/packages/virt-viewer/-/commits/11.0-2?ref_type=tags) instead of HEAD.

Thanks. It should be ok now.

PR #3139, opened 2023-10-19
upgpatch: v2ray

Xunop:
Resolved `go test` error; however, due to the current incompatibility of v2ray-core 5.7.0 with go version 1.21, the compilation is still failing.

This patch has been reported to the upstream as issue [v2ray/v2ray-core#2720](https://github.com/v2fly/v2ray-core/issues/2720)

### DIFF
--- a/v2ray/riscv64.patch
+++ b/v2ray/riscv64.patch
@@ -1,15 +1,22 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -12,12 +12,17 @@ license=('MIT')
+@@ -12,12 +12,31 @@ license=('MIT')
  depends=('glibc' 'v2ray-domain-list-community' 'v2ray-geoip')
  makedepends=('go' 'git')
  backup=(etc/v2ray/config.json)
 -source=("git+https://github.com/v2fly/v2ray-core.git#commit=$_commit")
 -sha512sums=('SKIP')
 +source=("git+https://github.com/v2fly/v2ray-core.git#commit=$_commit"
-+        "$pkgname-InitializeServerConfig-sleep.patch::https://github.com/v2fly/v2ray-core/pull/1786.patch")
++        "$pkgname-InitializeServerConfig-sleep.patch::https://github.com/v2fly/v2ray-core/pull/1786.patch"
++        "v2ray-testTCPConn2-timeout.patch"
++        "v2ray-testVMessKCP.patch"
++        "v2ray-kcpTest.patch")
++
 +sha512sums=('SKIP'
-+            '36553288b92bbf7d1ae65cc9efec95f1ca695f73e17804a72270e96382c685dea1dcae69ecfd9603bf7af0cf730080ac59c61c4b91cb724191e985032d5a5a03')
++            '36553288b92bbf7d1ae65cc9efec95f1ca695f73e17804a72270e96382c685dea1dcae69ecfd9603bf7af0cf730080ac59c61c4b91cb724191e985032d5a5a03'
++            'd681916026c2a224097c715cf21bfa53b0d61f496de828a38c94170d1a115081917e42cf3848f94ca76bdd5491be97c56f84cb6996584f204561f8f10796ce78'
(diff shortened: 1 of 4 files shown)


Xeonacid:
> Resolved `go test` error; however, due to the current incompatibility of v2ray-core 5.7.0 with go version 1.21, the compilation is still failing.

Thanks for the work! But do you mean it still does not compile with this updated patch? If so, I would suggest you try to fix it together in this PR. After all, a patch that doesn't work is not much meaningful. :P

> This patch has been reported to the upstream as issue [v2ray/v2ray-core#2720](https://github.com/v2fly/v2ray-core/issues/2720)

I'm wondering why do you create an issue with a patch instead of a PR. 🤔

Xunop:
> > Resolved `go test` error; however, due to the current incompatibility of v2ray-core 5.7.0 with go version 1.21, the compilation is still failing.
> 
> Thanks for the work! But do you mean it still does not compile with this updated patch? If so, I would suggest you try to fix it together in this PR. After all, a patch that doesn't work is not much meaningful. :P
> 
> > This patch has been reported to the upstream as issue [v2ray/v2ray-core#2720](https://github.com/v2fly/v2ray-core/issues/2720)
> 
> I'm wondering why do you create an issue with a patch instead of a PR. 🤔

emm, I've only fixed the problem with the code itself so far, the compilation error is because the upstream hasn't been updated to v5.8.0 .

You can check the logs [here](https://archriscv.felixc.at/.status/log.htm?url=logs/v2ray/v2ray-5.7.0-1.log), the error I fixed is the one in the logs. But since there is no update upstream (I updated locally to v5.8.0 for testing), so it compile fail. I started out just trying to fix the bug. But I realized that because of the incompatibility of the dependencies downloaded at compile time due to an upstream version change. Fortunately, the upstream has resolved the issue, I can I can fix the bug locally.

Sorry, my expressive ability is not particularly good, so if you have any questions please contact me right away.

Xeonacid:
> emm, I've only fixed the problem with the code itself so far, the compilation error is because the upstream hasn't been updated to v5.8.0 .

Then from archriscv side, just waiting for Arch to upgrade should be OK. Also we can make an Arch bug report that it's FTBFS with the new golang.

> You can check the logs [here](https://archriscv.felixc.at/.status/log.htm?url=logs/v2ray/v2ray-5.7.0-1.log), the error I fixed is the one in the logs. But since there is no update upstream (I updated locally to v5.8.0 for testing), so it compile fail. I started out just trying to fix the bug. But I realized that because of the incompatibility of the dependencies downloaded at compile time due to an upstream version change. Fortunately, the upstream has resolved the issue, I can I can fix the bug locally.

TBH I don't see any compile error from the log. Do you mean the failing check?

> Sorry, my expressive ability is not particularly good, so if you have any questions please contact me right away.

No worries at all! Actually you expressed clearly!

Xunop:
> > emm, I've only fixed the problem with the code itself so far, the compilation error is because the upstream hasn't been updated to v5.8.0 .
> 
> Then from archriscv side, just waiting for Arch to upgrade should be OK. Also we can make an Arch bug report that it's FTBFS with the new golang.
> 
> > You can check the logs [here](https://archriscv.felixc.at/.status/log.htm?url=logs/v2ray/v2ray-5.7.0-1.log), the error I fixed is the one in the logs. But since there is no update upstream (I updated locally to v5.8.0 for testing), so it compile fail. I started out just trying to fix the bug. But I realized that because of the incompatibility of the dependencies downloaded at compile time due to an upstream version change. Fortunately, the upstream has resolved the issue, I can I can fix the bug locally.
> 
> TBH I don't see any compile error from the log. Do you mean the failing check?

Yep, is the failing check. I just fix those errors. Since go was recently updated to 1.21, v2ay can't be compiled, although v2ray developers have solved this issue, but it's still in pre-realese. if the compilation succeeds and my patch is added, the check fail in the log I mentioned above will be solved. Actually, my patch only solves the check fail problem. So we have to wait for the upstream update, right?
> 
> > Sorry, my expressive ability is not particularly good, so if you have any questions please contact me right away.
> 
> No worries at all! Actually you expressed clearly!

Xeonacid:
> So we have to wait for the upstream update, right?

Yeah. And you can also create a bug report to Arch.

Xeonacid:
> Thank you for your patience. Do I need to close this pr?

No problem. You may leave it open and we can merge it after Arch upgrades.

PR #3140, opened 2023-10-19
upgpatch: luajit

aimixsaka:
- fix rotten
- update to upstream commit [172b660f2c0e75d64e85cd9b0df2ae14f1bde8eb](https://github.com/infiWang/LuaJIT/commit/172b660f2c0e75d64e85cd9b0df2ae14f1bde8eb)
- fetching source from github archive tarbar instead of repo commit

### DIFF
--- a/luajit/riscv64.patch
+++ b/luajit/riscv64.patch
@@ -1,24 +1,22 @@
-diff --git PKGBUILD PKGBUILD
-index 70afb4b7..141e9c88 100644
 --- PKGBUILD
 +++ PKGBUILD
-@@ -8,7 +8,7 @@
+@@ -7,7 +7,7 @@
+ 
  pkgname=luajit
- # LuaJIT has abandoned versioned releases and now advises using git HEAD
- # https://github.com/LuaJIT/LuaJIT/issues/665#issuecomment-784452583
--_commit=505e2c03de35e2718eef0d2d3660712e06dadf1f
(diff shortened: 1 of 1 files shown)


luajit/riscv64.patch
```diff
@@ -28,10 +26,27 @@ index 70afb4b7..141e9c88 100644
...
  
+-check() {
+-  cd "luajit-2.0-${_commit::7}"
+-
+-  # Make sure that _ct was updated
+-  test "${_ct}" == "$(cat .relver)"
+-}
++## FIXME
++## This check won't work in git+source,
++## as $pkgsrc is not a git repo
```
Xeonacid, line 42: Then we may switch back to source tarball instead of cloning from git. I used git for the first patch because we need a cherry-pick which is not the case now. :)

aimixsaka: Thanks. Can i use something like `https://github.com/infiWang/LuaJIT/archive/172b660f2c0e75d64e85cd9b0df2ae14f1bde8eb.zip`  ?

Xeonacid: > Can i use something like `https://github.com/infiWang/LuaJIT/archive/172b660f2c0e75d64e85cd9b0df2ae14f1bde8eb.zip` ?

`.tar.gz` is preferred. :)

aimixsaka: OK. :-)

aimixsaka:
Are there any other issues with this pr?  :-/

PR #3176, opened 2023-10-29
upgpatch: coin-or-cbc

Xunop:
Fix rotten.


r-value:
I think this patch can be removed since the config.guess has been updated upstream. No need to maintain it anymore.

PR #3202, opened 2023-10-31
addpatch: python-zict 2.2.0-2

hack3ric:
Backported map size fix from https://github.com/dask/zict/pull/111.

### DIFF
--- a/python-zict/riscv64.patch
+++ b/python-zict/riscv64.patch
@@ -1,0 +1,14 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -30,3 +30,11 @@ package() {
+   python setup.py install --prefix=/usr --root="${pkgdir}" --optimize=1 --skip-build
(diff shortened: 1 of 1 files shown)


python-zict/riscv64.patch
```diff
@@ -0,0 +1,14 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -30,3 +30,11 @@ package() {
+   python setup.py install --prefix=/usr --root="${pkgdir}" --optimize=1 --skip-build
+   install -Dm644 LICENSE.txt -t "${pkgdir}"/usr/share/licenses/${pkgname}/
+ }
++
++source+=(fix-map-size.patch::https://github.com/hack3ric/zict/commit/8aac36794415536b77a1b8318ddffe74278a758f.diff)
```
Xeonacid, line 8: Using `https://github.com/dask/zict/pull/111.diff` can result in a better tracking.

hack3ric: The pull request is targeted at 3.0.0 and upward, while the referenced commit is my own backport agains 2.2.0.

PR #3216, opened 2023-11-04
updpatch: rustypaste

aimixsaka:
- Fix the version of `ring` to `0.16.0` as there are multiple versions in Cargo.lock


aimixsaka:
Close as upstream of this package updated ring version to 0.17.5, there's no need to use patched ring.

PR #3218, opened 2023-11-04
rmvpatch:python-absl

XiePengHan:
Successfully built without applying patch.


Xeonacid:
You are creating a new empty patch while the file does not exist in our tree at all.

XiePengHan:
> You are creating a new empty patch while the file does not exist in our tree at all.



> You are creating a new empty patch while the file does not exist in our tree at all.

Thank you very much for your response. I believe I may have made a mistake in my approach

PR #3221, opened 2023-11-04
updpatch: deno 1.38.0-1

aimixsaka:
- Fix rotten
- Remove ring v0.16.20 patch as upstream is now using ring = "^0.17.0" (Successfully built without this)

### DIFF
--- a/deno/riscv64.patch
+++ b/deno/riscv64.patch
@@ -1,25 +1,17 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -11,7 +11,7 @@ url="https://deno.land"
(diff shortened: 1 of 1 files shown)


Xeonacid:
> * Update `makedepends`(add 'cmake' 'protobuf')

This is part of "Fix rotten". No need for a separate one next time to avoid any possible misunderstanding. Thanks. :)

aimixsaka:
> > * Update `makedepends`(add 'cmake' 'protobuf')
> 
> This is part of "Fix rotten". No need for a separate one next time to avoid any possible misunderstanding. Thanks. :)

OK. Thx. :-)

PR #3223, opened 2023-11-05
addpatch: nmon

XiePengHan:
- The downloaded file is a C file, so there's no need for `cd $srcdir`.
- The file being operated on is "lmon16p.c" not "nmon16p.c".
(I also suspected that the downloaded file should have been "nmon16p.c," but upon checking the official website, I discovered that they had discontinued the download path for "nmon16p.c." I attempted to find "nmon16p.c" through alternative means, but ultimately, I couldn't locate it. Therefore, I used "lmon16p.c.")

### DIFF
--- a/nmon/riscv64.patch
+++ b/nmon/riscv64.patch
@@ -1,0 +1,12 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -12,8 +12,7 @@ source=("https://downloads.sourceforge.net/$pkgname/lmon$pkgver.c")
+ sha256sums=('5dc6045f6725e3249a969918fae69663a1d669162087720babcdb90fce9e6b2a')
+ 
+ build() {
+-  cd "$srcdir"
+-  cc -o nmon nmon$pkgver.c $LDFLAGS $CFLAGS -g -O3 -lncurses -lm -D X86
++  cc -o nmon lmon$pkgver.c $LDFLAGS $CFLAGS -g -O3 -lncurses -lm -D X86
+ }
+ 
+ package() {


CoelacanthusHex:
> * The downloaded file is a C file, so there's no need for `cd $srcdir`.

This is a non-functional change, don't do it.

>     * The file being operated on is "lmon16p.c" not "nmon16p.c".
>       (I also suspected that the downloaded file should have been "nmon16p.c," but upon checking the official website, I discovered that they had discontinued the download path for "nmon16p.c." I attempted to find "nmon16p.c" through alternative means, but ultimately, I couldn't locate it. Therefore, I used "lmon16p.c.")

It should be reported to Arch Linux as FTBFS, rather than just changing the source downstream.

Xeonacid:
Besides, you just changed the filename in `build()`, but you should also change the source array.

nmon/riscv64.patch
```diff
@@ -0,0 +1,12 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -12,8 +12,7 @@ source=("https://downloads.sourceforge.net/$pkgname/lmon$pkgver.c")
+ sha256sums=('5dc6045f6725e3249a969918fae69663a1d669162087720babcdb90fce9e6b2a')
+ 
+ build() {
+-  cd "$srcdir"
+-  cc -o nmon nmon$pkgver.c $LDFLAGS $CFLAGS -g -O3 -lncurses -lm -D X86
++  cc -o nmon lmon$pkgver.c $LDFLAGS $CFLAGS -g -O3 -lncurses -lm -D X86
```
CoelacanthusHex, line 9: Are you sure this fixes the build on riscv64? You even still have the current architecture defined as x86.

CoelacanthusHex (changes requested):

nmon/riscv64.patch
XiePengHan: > Are you sure this fixes the build on riscv64? You even still have the current architecture defined as x86.

Thank you for your reminder.I'm confident that the test is valid. As a newcomer, I'm still relatively slow at reviewing source code. I understand that 'X86' is a preprocessing macro. What perplexes me is why there was a 404 error during our testing, while in your own testing, the URL was able to download successfully.

r-value:
Before any meaningful discussion, the patch you provided fails to apply to Arch Linux PKGBUILD.

nmon/riscv64.patch
r-value: The point is to port the package to riscv64 architecture, so you need to make sure that:
 + The changes you made is arch-specific
 + The artifacts in the package actually *works* on riscv64

PR #3229, opened 2023-11-06
addpkg: xdg-desktop-portal-wlr

XiePengHan:
The issue "Dependency 'gbm' not found" has been resolved.


r-value:
The error log complains about `gbm` instead of `gdm`. Your patch is probably a wrong approach.

XiePengHan:
> The error log complains about `gbm` instead of `gdm`. Your patch is probably a wrong approach.

 I made a mistake. I will try to fix it again tomorrow.

XiePengHan:
> Collaborator

Thank you for your guidance.

PR #3248, opened 2023-11-08
addpatch: vue-language-server 1.8.22-1

phanen:
Increase timeout to pass test in `check()`.
Upstream: https://github.com/vuejs/language-tools/pull/3721.


Xeonacid:
For decreasing repo size, we don't need to store the patch if it has a source url. :)

PR #3255, opened 2023-11-12
updpatch: update to linux 6.6.1.arch1

CoelacanthusHex:
- enable some VisionFive2 driver
- drop old disable-sv57 patch because 6.6 include better way


linux/riscv64.config-patch
```diff
@@ -773,15 +780,23 @@
...
 +# CONFIG_ARM_SUN8I_A33_MBUS_DEVFREQ is not set
  CONFIG_PM_DEVFREQ_EVENT=y
  CONFIG_EXTCON=y
  
-@@ -9057,8 +9262,10 @@
+@@ -8609,6 +8820,7 @@
+ CONFIG_PALMAS_GPADC=m
+ CONFIG_RICHTEK_RTQ6056=m
+ CONFIG_SD_ADC_MODULATOR=m
++# CONFIG_SUN20I_GPADC is not set
```
felixonmars, line 795: Looks like this is a driver for D1, should we enable it?

linux/riscv64.patch
```diff
@@ -1,87 +0,0 @@
-diff --git PKGBUILD PKGBUILD
-index 2b0f6d9..b1b9999 100644
---- PKGBUILD
-+++ PKGBUILD
-@@ -31,6 +31,8 @@ source=(
-   https://cdn.kernel.org/pub/linux/kernel/v${pkgver%%.*}.x/${_srcname}.tar.{xz,sign}
-   $url/releases/download/$_srctag/linux-$_srctag.patch.zst{,.sig}
-   config  # the main kernel config file
-+  riscv64.config-patch
```
felixonmars, line 9: I assume the other changes are still needed...

felixonmars (changes requested):

linux/riscv64.config-patch
CoelacanthusHex: done

linux/riscv64.patch
CoelacanthusHex: done

PR #3268, opened 2023-11-17
updpatch: refind 0.14.0.2-1

phanen:
Add a missing flag to build `refind.efi`.


r-value:
are you sure this works on risc-v?

phanen:
Not sure... but efi cannot be built without this flag in latest release, since refind hasn't offcially support riscv64.

I just find a long patch in https://github.com/loongarchlinux/extra/blob/f96c8a0f832c88d39134c8d38e2ab61b43235cd7/refind/refind-la64-0.14.0.patch#L522, maybe it works for 0.14.0.

PR #3269, opened 2023-11-17
init blacklist for machines with thead cpu

kxxt:
thead-blacklist.txt applies to all machines with thead cpu.


kxxt:
This PR has been updated for current situation.

PR #3278, opened 2023-11-18
addpatch: rocm-smi-lib

aimixsaka:
- remove unneeded c++ flag `-m64` and unsupported flags `-msse -msse2` for riscv64

### DIFF
--- a/rocm-smi-lib/riscv64.patch
+++ b/rocm-smi-lib/riscv64.patch
@@ -1,0 +1,24 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -11,11 +11,19 @@ license=('custom:NCSAOSL')
+ depends=('hsa-rocr' 'python')
+ makedepends=('cmake')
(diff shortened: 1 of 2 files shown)


Xeonacid:
Backporting https://github.com/RadeonOpenCompute/rocm_smi_lib/commit/1cf05dd9c704f0fad0f535330ee1955914a4fe86 should be better.

aimixsaka:
> Backporting [RadeonOpenCompute/rocm_smi_lib@1cf05dd](https://github.com/RadeonOpenCompute/rocm_smi_lib/commit/1cf05dd9c704f0fad0f535330ee1955914a4fe86) should be better.

Thanks. Will do. :)

aimixsaka:
Re-pushed patch, which backport RadeonOpenCompute/rocm_smi_lib@1cf05dd, to remove '-m64 -msse -msse2' c++ flags on architectures that don't support them. 

This prevent breaking on x86_64 system compared to just sed flags.

PR #3279, opened 2023-11-18
addpatch: rocminfo 5.7.1-1

aimixsaka:
- Remove unneeded c++ flag `-m64` for riscv64, see riscv-collab/riscv-gnu-toolchain#670


r-value:
RadeonOpenCompute/rocminfo#63

aimixsaka:
Re-pushed patch, which applied patch RadeonOpenCompute/rocminfo#63 to enable `-m64` flag only on x86_64 host.

This prevent breaking on x86_64 system compared to just sed flags.

PR #3281, opened 2023-11-18
qemu-user-blacklist: git

r-value:
check fails in qemu-user


r-value:
superseded by #3284，closing

PR #3283, opened 2023-11-19
addpatch: skia-sharp

kxxt:
Use system gn & ninja.

### DIFF
--- a/skia-sharp/riscv64.patch
+++ b/skia-sharp/riscv64.patch
@@ -1,0 +1,32 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -43,6 +43,8 @@ pkgver() {
+   git describe --tags | sed 's/^v//; s/-preview.*//'
+ }
+ 
++makedepends+=(ninja gn)
++
+ prepare() {
+   cd SkiaSharp
+   for submodule in externals/{depot_tools,skia}; do
+@@ -75,9 +77,9 @@ build(){
+     extra_ldflags+="\"${ldflag}\","
+   done
+ 
+-  bin/gn gen out \
++  /usr/bin/gn gen out \
+     --args="target_os=\"linux\" \
+-            target_cpu=\"x64\" \
++            target_cpu=\"riscv64\" \
+             cc=\"clang\" \
+             cxx=\"clang++\" \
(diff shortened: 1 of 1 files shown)


Xeonacid:
I'm wondering why Arch don't use the system ones? (It seems intentional since they set $PATH.)

kxxt:
> I'm wondering why Arch don't use the system ones? (It seems intentional since they set $PATH.)

Previously this package has the following mark:

> The package require the Google internal gn executable. We need to downloaded it from it's internal compile server but there is not RISC-V64 compatible binary on the download server.  They make some modification in the executable so the extra/gn is not usable for this package.

I guess that's the reason. But now oss gn can setup the build without trouble. 

And the changes in this patch does not fit into an Arch bug report because there's no FTBFS on x86_64 side. No need to bother the package maintainers.

Xeonacid:
> And the changes in this patch does not fit into an Arch bug report because there's no FTBFS on x86_64 side. No need to bother the package maintainers.

Actually I don't think so. Using system makedepends should be an enhancement.

kxxt:
> > And the changes in this patch does not fit into an Arch bug report because there's no FTBFS on x86_64 side. No need to bother the package maintainers.
> 
> Actually I don't think so. Using system makedepends should be an enhancement.

Yes, it is an enhancement **that doesn't affect the build output**. I don't think it deserve a **bug** report. It is neither a bug nor a feature request. We could submit these things when Arch opens the merge request feature for packages on their gitlab.

r-value:
Actually I think it *is* a bug -- it's wasting user's disk space. It deserves a report to inform Arch team or at least be aware of the package maintainer's intention. It looks odd so there must be a strong reason to do so.

kxxt:
> Actually I think it _is_ a bug -- it's wasting user's disk space. It deserves a report to inform Arch team or at least be aware of the package maintainer's intention. It looks odd so there must be a strong reason to do so.

Actually adding gn and ninja to makedepends takes extra disk space. The prebuilt version is always there unless we go out of our way to patch depot_tools. Feel free to submit a ticket if you insist :)

felixonmars:
I have mixed feelings here because the current x86_64 approach does look disgusting to me. They may have their reasoning though :/

PR #3291, opened 2023-11-21
addpatch: buildbot

kxxt:
- Increase test timeout of `start_test` by backporting upstream commit. We can probably remove this patch at the next release.
- Add to qemu blacklist because chromium can't execute in qemu: https://www.mail-archive.com/qemu-devel@nongnu.org/msg803994.html


yan12125:
This patch is already in upstream 3.10.0, resulting in patch failures: https://archriscv.felixc.at/.status/log.htm?url=logs/buildbot/buildbot-3.10.0-2.log

PR #3308, opened 2023-12-02
addpatch: pcre2

felixonmars:
Apply a patch to fix sv48/sv57 crashes. Upstream hasn't confirmed the
fix [1] yet but at least the tests are passing now.

Thanks to @ksco !

[1] https://github.com/zherczeg/sljit/pull/223


felixonmars:
I'll follow the updates on the sljit PR because there is still a failing test.

PR #3335, opened 2023-12-10
addpatch: snapshot

Xunop:
Fix `Cargo clippy` timeout error, upstreamed to snapshot: https://gitlab.gnome.org/GNOME/snapshot/-/issues/111


Xunop:
archlinux has update PKGBUILD, I will update this patch.

Xunop:
patch updated.

PR #3378, opened 2023-12-26
updpatch: ldc 3:1.35.0-1.1

hack3ric:
...and here comes the second part:
- Switch to LLVM 16, in line with Arch Linux.
- Remove all but one atomic hack. This is needed because the tests somehow requires double word atomic operations, which is not supported on riscv64.


hack3ric:
Please merge it when LDC compiles successfully with the first patch.

PR #3384, opened 2023-12-27
addpatch: tilix 1.9.6-1

hack3ric:
LTO failed (both full and thin):

```
/usr/bin/ld: /tmp/lto-llvm-ea1ef0.o: can't link soft-float modules with double-float modules
/usr/bin/ld: failed to merge target specific data of file /tmp/lto-llvm-ea1ef0.o
```


hack3ric:
It seems [LDC does produce double-precision object by default](https://github.com/ldc-developers/ldc/pull/4474), so it's more likely an LTO problem.

PR #3389, opened 2023-12-28
addpatch: imapsync

Xunop:
- Using other test scripts, ref: https://github.com/imapsync/imapsync/issues/431
- Comment out some tests because IPv6 needs to be used
- Upstreamed to archlinux: https://gitlab.archlinux.org/archlinux/packaging/packages/imapsync/-/merge_requests/1

### DIFF
--- a/imapsync/fix-test.patch
+++ b/imapsync/fix-test.patch
@@ -1,0 +1,47 @@
+diff --git a/imapsync b/imapsync
+index e842961..8ad50b9 100644
+--- a/imapsync
++++ b/imapsync
+@@ -7776,7 +7776,7 @@ sub tests_mailimapclient_connect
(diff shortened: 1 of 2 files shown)


felixonmars:
> Comment out some tests because IPv6 needs to be used

I would somehow prefer to avoid these changes and mark them as ipv6-needed in some way in my build scheduler (since we do have ipv6-enabled builders, currently qemu-user-only, though).

Xunop:
> I would somehow prefer to avoid these changes and mark them as ipv6-needed in some way in my build scheduler (since we do have ipv6-enabled builders, currently qemu-user-only, though).

Happy New Year! ArchLinux x86 has updated this package, so we no longer need this patch. Just run it in qemu-user for IPv6.

PR #3399, opened 2024-01-01
addpatch: devtools 1:1.0.4-1

magic0whi:
Since Arch Linux RISC-V is using a different style of repo link. The correspondence `sed` regex in `arch-nspawn` also needs to be refit: from
  http://riscv.mirror.pkgbuild.com/repo/extra,
to
  http://riscv.mirror.pkgbuild.com/repo/$repo.

This fix sequence 404 errors while retrieving any package in clean chroot due to that `/etc/pacman.d/mirrorlist` still keep the form of 'Server = http[s]://**/repo/extra'.

At this time, any package build try in a claen chroots will result in fail if it has any make dependent packages not in group 'base-devel'.

### DIFF
--- a/devtools-riscv64/PKGBUILD
+++ b/devtools-riscv64/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=devtools-riscv64
 epoch=1
-pkgver=1.0.4+patch1
+pkgver=1.0.4+patch2
 pkgrel=1
 pkgdesc='Tools for Arch Linux RISC-V package maintainers'
 arch=('x86_64' 'riscv64')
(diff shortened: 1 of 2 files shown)


magic0whi:
TODO: Modify `PKGBUILD` (I forgot it)

Xeonacid:
> At this time, any package build try in a claen chroots will result in fail if it have any make dependent packages not in group 'base-devel'.

Could you leave a reproduction step? We build a lot of packages using `extra-riscv64-build` on qemu-user or real hardware, but have not seen such problem. 🤔

magic0whi:
> > At this time, any package build try in a claen chroots will result in fail if it have any make dependent packages not in group 'base-devel'.
> 
> Could you leave a reproduction step? We build a lot of packages using `extra-riscv64-build` on qemu-user or real hardware, but have not seen such problem. 🤔

Okey, I’ll pick a package to build, and then check out the different. My assume is that the `extra-riscv64-build` either prepared all the make depends at chroot initialization or it first checkd out makedeps in PKGBUILD and called pacman with host mirrorlist file so at the time makepkg in chroot being called it does not need to install anymore packages.

magic0whi:
> > > At this time, any package build try in a claen chroots will result in fail if it have any make dependent packages not in group 'base-devel'.
> > 
> > 
> > Could you leave a reproduction step? We build a lot of packages using `extra-riscv64-build` on qemu-user or real hardware, but have not seen such problem. 🤔
> 
> Okey, I’ll pick a package to build, and then check out the different. My assume is that the `extra-riscv64-build` either prepared all the make depends at chroot initialization or itself checkd out makedeps in PKGBUILD and called pacman using host mirrorlist file at early time so when makepkg in chroot being excuted it does not need to install anymore packages.

Sorry for the inconvenience. Just found out that I mentioned I uses the ‘makechrootpkg -c -r $CHROOT’ to build packages (which is described as the classic way, just because I wanna pass some the parameter “-A” to ignore the arch limit in PKGBUILD)

CoelacanthusHex:
> > > > At this time, any package build try in a claen chroots will result in fail if it have any make dependent packages not in group 'base-devel'.
> > > 
> > > 
> > > Could you leave a reproduction step? We build a lot of packages using `extra-riscv64-build` on qemu-user or real hardware, but have not seen such problem. 🤔
> > 
> > 
> > Okey, I’ll pick a package to build, and then check out the different. My assume is that the `extra-riscv64-build` either prepared all the make depends at chroot initialization or itself checkd out makedeps in PKGBUILD and called pacman using host mirrorlist file at early time so when makepkg in chroot being excuted it does not need to install anymore packages.
> 
> Sorry for the inconvenience. Just found out that I’ve never mentioned I uses the ‘makechrootpkg -c -r $CHROOT’ to build packages (which is described as the classic way, just because I wanna pass some parameters like “-A” which ignore the arch limit in PKGBUILD)

You should change `arch` array in PKGBUILD instead. Invoking makechrootpkg manually is unsupported way in nowadays Arch packaging.

PR #3418, opened 2024-01-08
updpatch: arch-audit 0.1.20-2

phanen:
Fix rotten.

### DIFF
--- a/arch-audit/riscv64.patch
+++ b/arch-audit/riscv64.patch
@@ -1,10 +1,6 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -15,9 +15,13 @@ backup=('etc/arch-audit/settings.toml')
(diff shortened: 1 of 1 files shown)


arch-audit/riscv64.patch
```diff
@@ -14,3 +10,11 @@
    cargo build --release --locked
  }
  
+@@ -31,7 +34,6 @@ package() {
+   cd "${pkgname}-${pkgver}"
+ 
+   make DESTDIR="${pkgdir}" PREFIX=/usr install
+-
```
Xeonacid, line 17: Please don't touch this. :)

Xeonacid (changes requested):

arch-audit/riscv64.patch
phanen: Fixed (:

PR #3456, opened 2024-01-21
updpatch: linux-th1520 6.7.revyos6

CoelacanthusHex:
- enable RPMSG_THEAD_LIGHT, it must set RPMSG and RPMSG_VIRTIO to yes
  because RPMSG_THEAD_LIGHT build failed as module.
- LIGHT_PMIC_WATCHDOG=y because it can't be loaded as module.
- PINCTRL_TH1520=y because uart depends on it.


RevySR:
https://github.com/revyos/th1520-linux-kernel/issues/5

CONFIG_PINCTRL_TH1520 should be builtin. Because uart depends on it.

RevySR:
LIGHT_PMIC_WATCHDOG should be builtin. 
This driver does not have the ability to be automatically loaded as a module.

PR #3486, opened 2024-01-30
updpatch: java11-openjdk 11.0.22.u7-1

hack3ric:
Use updated Java 11 patch from openSUSE.


hack3ric:
Can confirm this fixes shatter-pixel-dungeon (and possibly LibreOffice and a bunch of Java packages).

PR #3521, opened 2024-02-11
addpatch: typescript 5.3.3-1

hack3ric:
Remove dprint from TypeScript's dependencies and building processes since dprint itself and its native plugins (non-WebAssembly ones) relies on prebuilt binaries.

Attempt to enable dprint using https://github.com/dprint/dprint/pull/820 and disable its native Prettier plugins (only used to format YAML files) did not succeed, as [dprint-plugin-typescript (a WebAssembly plugin) produced inconsistent typescript.d.ts between riscv64 and x86_64](https://gist.github.com/hack3ric/bd7dd2364cadba4b01d6c834396ad06f). It is probably caused by immature Cranelift JIT implementation on RISC-V. The difference is only cosmetic though, therefore we can safely ignore the test for now.

### DIFF
--- a/typescript/riscv64.patch
+++ b/typescript/riscv64.patch
@@ -1,0 +1,35 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -12,11 +12,30 @@ url=http://www.typescriptlang.org
+ license=('Apache')
+ depends=('nodejs')
+ makedepends=('git' 'npm' 'rsync')
+-source=("git+https://github.com/microsoft/$_name.git#tag=v$pkgver")
+-b2sums=('SKIP')
++source=("git+https://github.com/microsoft/$_name.git#tag=v$pkgver"
++        "remove-dprint.patch"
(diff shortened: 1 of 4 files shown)


typescript/riscv64.patch
```diff
@@ -0,0 +1,37 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -11,12 +11,31 @@ arch=('any')
+ url=http://www.typescriptlang.org
+ license=('Apache')
+ depends=('nodejs')
+-makedepends=('git' 'npm' 'rsync')
+-source=("git+https://github.com/microsoft/$_name.git#tag=v$pkgver")
+-b2sums=('SKIP')
++makedepends=('git' 'npm' 'rsync' 'cargo')
```
Xeonacid, line 10: Why do we need cargo here?

hack3ric: Left when I tested dprint. Removed now.

PR #3546, opened 2024-02-21
updpatch: luajit 2.1.1702376626-1

hack3ric:
- Specify how to choose commit from https://github.com/infiWang/LuaJIT. See diff for detail.
- Make package version not necessarily align with Arch x86_64, for RISC-V-related commits need to be included.

### DIFF
--- a/luajit/riscv64.patch
+++ b/luajit/riscv64.patch
@@ -1,49 +1,31 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -7,9 +7,9 @@
+@@ -7,9 +7,11 @@
  
  pkgname=luajit
  # LuaJIT has a "rolling release" where you should follow git HEAD
--_commit=e826d0c101d750fac8334d71e221c50d8dbe236c
-+_commit=4da850743b197a7e864ab811ec3dce7204040a7b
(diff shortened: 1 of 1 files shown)


luajit/riscv64.patch
```diff
@@ -1,49 +1,32 @@
...
+@@ -7,9 +7,12 @@
  
  pkgname=luajit
  # LuaJIT has a "rolling release" where you should follow git HEAD
--_commit=e826d0c101d750fac8334d71e221c50d8dbe236c
-+_commit=4da850743b197a7e864ab811ec3dce7204040a7b
+-_commit=ff204d0350575cf710f6f4af982db146cb454e1a
++# note to riscv: choose commit close to (but newer than) Arch upstream from `riscv` branch of infiWang/LuaJIT
++# also check `riscv-stub` branch so the selected commit contains necessary changes to RISC-V
++# make sure to also update $_ct below
```
Xeonacid, line 10: I don't fully understand here. IIUC, "choose commit close to (but newer than) Arch upstream from `riscv` branch of infiWang/LuaJIT" is to follow the commit of Arch. But how can `riscv-stub` branch help? If we need to check for new riscv port commit, why not examine `riscv` branch tree directly?

PR #3587, opened 2024-03-09
updpatch: opus 1.5.1-1

felixonmars:


felixonmars:
Requires `nocheck` for now because both unmatched & sg2042 timed out. A proper solution should be provided later.

PR #3630, opened 2024-03-22
addpatch: cloudflared 2024.1.5-1

hack3ric:
- ~~Drop `-extldflags` since Go linker on riscv64 does not support most of it~~
- Fix LDFLAGS issue. Upstreamed to https://gitlab.archlinux.org/archlinux/packaging/packages/cloudflared/-/merge_requests/1.
- Remove unsupported `-race` from go test

### DIFF
--- a/cloudflared/riscv64.patch
+++ b/cloudflared/riscv64.patch
@@ -1,0 +1,20 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -54,7 +54,7 @@ build() {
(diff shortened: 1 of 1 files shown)


Xeonacid:
> Drop -extldflags since Go linker on riscv64 does not support most of it

Which flag is unsupported on riscv? i guess it fails to build because `${LDFLAGS}` has spaces in it, but is not surrounded by quotation marks.

PR #3805, opened 2024-05-28
updpatch: qt6-webengine 6.7.1-1

kxxt:
Update patch so that there's only one prepare function. (Arch added prepare())


kxxt:
Patch updated to base on 6.7.1-1 since 6.7.1-3 is still in extra-staging

PR #3845, opened 2024-06-18
addpkg: sharutils

qyl27:
Will still get a permission denied issue when packaging, but it does not affect its outputted result.  
That issue can be reproduced in x86 packaging, reported in https://gitlab.archlinux.org/archlinux/packaging/packages/sharutils/-/issues/2.  
Outdated config.guess issue was reported in https://savannah.gnu.org/support/index.php?111077.

### DIFF
--- a/sharutils/riscv64.patch
+++ b/sharutils/riscv64.patch
@@ -1,15 +1,11 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -25,10 +25,12 @@ build() {
+@@ -19,6 +19,8 @@ sha256sums=('2b05cff7de5d7b646dc1669bc36c35fdac02ac6ae4b6c19cb3340d87ec553a9a'
+ prepare() {
  	cd "${srcdir}/${pkgname}-${pkgver}"
- 	CFLAGS+=' -fcommon' # https://wiki.gentoo.org/wiki/Gcc_10_porting_notes/fno_common
- 	CFLAGS+=' -Wno-format-security' # fix build with gettext function
-+	autoreconf -fi
- 	./configure \
- 		--prefix=/usr \
- 		--mandir=/usr/share/man \
- 		--infodir=/usr/share/info \
-+		--disable-dependency-tracking
- 
- 	make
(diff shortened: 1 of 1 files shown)


sharutils/riscv64.patch
```diff
@@ -0,0 +1,15 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -25,10 +25,12 @@ build() {
+ 	cd "${srcdir}/${pkgname}-${pkgver}"
+ 	CFLAGS+=' -fcommon' # https://wiki.gentoo.org/wiki/Gcc_10_porting_notes/fno_common
+ 	CFLAGS+=' -Wno-format-security' # fix build with gettext function
++	autoreconf -fi
```
CoelacanthusHex, line 7: Should be in `prepare()` function.

sharutils/riscv64.patch
```diff
@@ -0,0 +1,15 @@
...
+ 	./configure \
+ 		--prefix=/usr \
+ 		--mandir=/usr/share/man \
+ 		--infodir=/usr/share/info \
++		--disable-dependency-tracking
```
CoelacanthusHex, line 12: Need a reason why we need it and why it's RISC-V specific.

CoelacanthusHex (commented):
Lack the upstream report link on outdated config.guess.

qyl27:
There is an error during building after `autoreconf`, it is an invalid argument of touch, after analyzing the output of automake, it is caused by the script selecting `touch` as the alternative to `autogen`.  
In archlinux upstream, they used the `config.guess` and `config.sub` which are in the source pack and outdated, that did not happenned, but after `autoreconf`, it appears due to enabling automatic dependency tracking by default.  
This issue was reported to upstream, see: https://savannah.gnu.org/support/index.php?111078 .  
(Thanks to @CoelacanthusHex , she is a good teacher.)

qyl27:
The permission denied issue was fixed in makepkg.  
https://gitlab.archlinux.org/archlinux/packaging/packages/sharutils/-/issues/2 .

CoelacanthusHex:
Please squash commits and update commit message to include what and why you did them, and add package version in the commit and PR title.

PR #3864, opened 2024-07-06
addpatch: festival 2.5.0-5

qyl27:
Outdated `config.guess` and `config.sub` in `festival` and `speech_tools` were reported to Arch Linux.  
The mail list of upstream seems down for several years. (Original upstream: http://festvox.org/maillists.html, I have tried to join and send mail.)
Issue: 
https://github.com/festvox/festival/issues/74
https://github.com/festvox/speech_tools/issues/56

### DIFF
--- a/festival/riscv64.patch
+++ b/festival/riscv64.patch
@@ -1,0 +1,16 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -42,6 +42,13 @@ prepare() {
+ 
+   # Avoid make failure on making scripts and docs
+   sed -i "s|examples bin doc|examples|" festival/Makefile
++
++  cd speech_tools
++  autoreconf -fi
++  cd ..
++  cd festival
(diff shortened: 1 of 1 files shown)


Xeonacid:
> Outdated `config.guess` and `config.sub` in `festival` and `speech_tools` were reported to Arch Linux. 

This should not be reported to Arch Linux as it does not affect x86_64 and the outdated files are not distro's fault. Only upstream should be notified. :)

qyl27:
> > Outdated `config.guess` and `config.sub` in `festival` and `speech_tools` were reported to Arch Linux.
> 
> This should not be reported to Arch Linux as it does not affect x86_64 and the outdated files are not distro's fault. Only upstream should be notified. :)

Yes, I have no idea about how to get contact to upstream for now. Their mail list seems not available which is the only way mentioned on their website.

PR #3877, opened 2024-07-14
fixpatch: valgrind

Nightwulf:
updated valgrind commit to base on

### DIFF
--- a/fuseiso/riscv64.patch
+++ b/fuseiso/riscv64.patch
@@ -1,0 +1,15 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -21,6 +21,12 @@ sha256sums=('8b242e077d66cd20900c59c905ff90b4c934b0613dd5a20facb0b1260ac5fd88'
+             'def5dd8dc20ad44ec659c20472744db6329684fc1d53f451e28bae655c799a8f'
+             '520e01cd544d21e6400c5d63bcaf57e344fbb59406a6a61636dc0d4cb6447bd1')
+ 
++prepare() {
++  cd ${pkgname}-${pkgver}
++  cp /usr/share/autoconf/build-aux/config.guess config.guess
++  cp /usr/share/autoconf/build-aux/config.sub config.sub
++}
++
+ build() {
+   cd ${pkgname}-${pkgver}
+ 


Nightwulf:
sure, done...thx for the hint!

valgrind/riscv64.patch
```diff
@@ -1,63 +1,48 @@
---- PKGBUILD
-+++ PKGBUILD
-@@ -20,47 +20,35 @@ arch=('x86_64')
+--- PKGBUILD	2024-07-14 14:56:50.249389360 +0200
++++ PKGBUILD.rv64	2024-07-14 16:52:37.578265985 +0200
```
felixonmars, line 2: Hi, please don't change the filename :)

valgrind/riscv64.patch
```diff
@@ -1,63 +1,48 @@
...
-+  patch -Np1 < ../fix-riscv64-fp-ite-csel.patch
- 
-   autoreconf -ifv
+@@ -118,3 +110,6 @@
  }
+ 
+ # vim: ts=2 sw=2 et:
++sha256sums=('976fb5e53c6121a29bb8421e61926911c1afc4cae8292c634135e5cbc3619fde'
++            '63a0b914fc66decf25760523eb7d7031bf08bf9e7e26f891b03b9e72c375d590'
++            '4e6be3a1799c17a6e843ab1966a3a68ac0ffa83d4b2230ce1b607518c42a31a2')
```
felixonmars, line 48: I'll still prefer using the original means of checksums so those unmodified patches won't cause any surprises.

Nightwulf: The filename is a stupid mistake on my side...using it when a patch don't run at first to test it locally...shouldn't have made it's way to the patch...will change that. Sorry. 
The checksums have been calculated by updpkgsums. But since I could resolve it without a new patch I wonder why they are different. I'll have another look at it tomorrow.
Thx for your remarks!

Nightwulf: Had another look....the changed checksums result from the patch using a newer commit of the project. The patch before this PR hat the same lines changed.

felixonmars: No I actually meant that the sha256sums added at the end of the PKGBUILD were not present in either Arch upstream or the original patch here.

PR #3893, opened 2024-07-16
updpatch: ltrace 0.7.3-5

hack3ric:
- Backport https://gitlab.com/cespedes/ltrace/-/merge_requests/21
- Fix missing includes and function arguments

### DIFF
--- a/ltrace/riscv64.patch
+++ b/ltrace/riscv64.patch
@@ -1,21 +1,29 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -12,8 +12,16 @@ license=('GPL')
- depends=('elfutils' 'libunwind')
- makedepends=('dejagnu')
- backup=('etc/ltrace.conf')
--source=(https://sources.archlinux.org/other/$pkgname/$pkgname-$pkgver.tar.bz2)
--sha256sums=('0e6f8c077471b544c06def7192d983861ad2f8688dd5504beae62f0c5f5b9503')
-+source=(https://sources.archlinux.org/other/$pkgname/$pkgname-$pkgver.tar.bz2
-+        "riscv64-support.patch")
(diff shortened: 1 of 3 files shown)


ltrace/riscv64.patch
```diff
@@ -1,21 +1,29 @@
...
  build() {
    cd "$srcdir/$pkgname-$pkgver"
+ 
+@@ -37,4 +45,11 @@ package(){
+   make DESTDIR="$pkgdir" install
+ }
+ 
++source+=(riscv64-support.patch
++         support-c-ebreak.patch::https://gitlab.com/cespedes/ltrace/-/merge_requests/21.diff
++         fix-includes.patch)
```
Xeonacid, line 24: Since this PR has not been merged and there may be future changes to the checksum, and we still need to patch it, how about saving the (patched) patch in our repo instead of referencing the upstream URL?
Additionally, it would be best to report this `fix-includes.patch` to the PR author.

ltrace/riscv64.patch
```diff
@@ -1,21 +1,29 @@
...
++sha256sums+=('3261e93f9ae91982afd382046de2f2a7399402c74981724c53cb8b4e25745beb'
++             'e6d7a6e7c940d534195e62312acc19d2c2be3012ba1c8ae8e82a6f088ff2ede6'
++             'SKIP')
```
Xeonacid, line 27: Why is the`fix-includes.patch` checksum skipped?

ltrace/riscv64.patch
hack3ric: It's already fixed in the `main` branch.

ltrace/riscv64.patch
hack3ric: Oops.

PR #3907, opened 2024-07-18
updpatch: highway 1.2.0-1

qyl27:
Backport https://github.com/google/highway/commit/c95cc0237d2f7a0f5ca5dc3fb4b5961b2b1dcdfc into 1.2.0 to build.

### DIFF
--- a/highway/riscv64.patch
+++ b/highway/riscv64.patch
@@ -1,8 +1,24 @@
-diff --git PKGBUILD PKGBUILD
-index d59be29..9c383f2 100644
 --- PKGBUILD
 +++ PKGBUILD
-@@ -19,6 +19,7 @@ build() {
+@@ -9,8 +9,15 @@ url='https://github.com/google/highway/'
+ license=('Apache-2.0' 'BSD-3-Clause')
+ depends=('gcc-libs')
+ makedepends=('cmake' 'gtest')
+-source=("https://github.com/google/highway/archive/${pkgver}/${pkgname}-${pkgver}.tar.gz")
+-sha256sums=('7e0be78b8318e8bdbf6fa545d2ecb4c90f947df03f7aadc42c1967f019e63343')
++source=("https://github.com/google/highway/archive/${pkgver}/${pkgname}-${pkgver}.tar.gz"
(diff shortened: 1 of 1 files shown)


highway/riscv64.patch
```diff
@@ -1,8 +1,24 @@
...
 +++ PKGBUILD
-@@ -19,6 +19,7 @@ build() {
+@@ -9,8 +9,15 @@ url='https://github.com/google/highway/'
+ license=('Apache-2.0' 'BSD-3-Clause')
+ depends=('gcc-libs')
+ makedepends=('cmake' 'gtest')
+-source=("https://github.com/google/highway/archive/${pkgver}/${pkgname}-${pkgver}.tar.gz")
+-sha256sums=('7e0be78b8318e8bdbf6fa545d2ecb4c90f947df03f7aadc42c1967f019e63343')
++source=("https://github.com/google/highway/archive/${pkgver}/${pkgname}-${pkgver}.tar.gz"
++        "c95cc0237d2f7a0f5ca5dc3fb4b5961b2b1dcdfc.patch")
```
Deriveda, line 10: Could we ref the patch by URL instead of storing it? Also, it should have a meaningful name like `highway-disable-RVV-runtime-dispatch.patch` or sth else.

highway/riscv64.patch
```diff
@@ -1,8 +1,24 @@
...
++sha256sums=('7e0be78b8318e8bdbf6fa545d2ecb4c90f947df03f7aadc42c1967f019e63343'
++            'SKIP')
```
Deriveda, line 12: Why is the checksum skipped?

highway/riscv64.patch
Xeonacid, line 10: Could we ref the patch by URL instead of storing it? Also, it should have a meaningful name like `highway-disable-RVV-runtime-dispatch.patch` or sth else.

highway/riscv64.patch
Xeonacid, line 12: Why is the checksum skipped?

Xeonacid (changes requested):

Xeonacid:
> Also, it should have a meaningful name like highway-disable-RVV-runtime-dispatch.patch or sth else.

This seems unresolved?

Xeonacid (approved):

PR #3917, opened 2024-07-21
addpatch: libraw1394

Nightwulf:


Xeonacid:
May you try to contact upstream to let them update the relevant files so we don't need to always patch here? Thanks!

Nightwulf:
That has already been done 2 days ago. Since it is blocking ffmpeg from updating, which then blocks other packages, we should perhaps keep an eye on this bug but despite that live with that patch currently.

felixonmars:
https://bugzilla.kernel.org/show_bug.cgi?id=219066 for the record

PR #3928, opened 2024-07-31
rmvpatch: nodejs-lts-{iron,hydrogen}

kxxt:
Remove old workaround.

Depends on #3927


kxxt:
I think the nodejs package also needs to be rebuilt with the new gcc. The workaround for nodejs package is removed in bd8eef0a114f37de18b6753a525749ff504bb850. https://archriscv.felixc.at/.status/log.htm?url=logs/hedgedoc/hedgedoc-1.9.9-2.log is possibly the same bug.

PR #3942, opened 2024-08-03
updpatch: ldc 3:1.39.0-1

hack3ric:
LLVM 18 rebuild; also built successfully using lld so re-enable it as well.


wszqkzqk:
Hello! Their is an upstream fix to link atomic on riscv and loongarch for druntime, would you like to try: https://github.com/ldc-developers/ldc/pull/4864

hack3ric:
> Hello! Their is an upstream fix to link atomic on riscv and loongarch for druntime, would you like to try: https://github.com/ldc-developers/ldc/pull/4864

Sure!

PR #3980, opened 2024-08-11
addpatch: nominatim-ui 3.5.3-1

qyl27:
Fix the check of nominatim-ui.  
The test code needs to launch a chromium which cannot run on qemu-user.  
The timeout is too short for RISC-V devices with extremely low performance, it might be discussed at https://github.com/osm-search/nominatim-ui/issues/253 .

### DIFF
--- a/nominatim-ui/riscv64.patch
+++ b/nominatim-ui/riscv64.patch
@@ -1,0 +1,15 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -14,9 +14,10 @@ backup=('etc/webapps/nominatim-ui/config.theme.js')
+ install=nominatim-ui.install
+ source=("${pkgname}-${pkgver}.tar.gz::${url}/archive/tags/v${pkgver}.tar.gz"
+         'nginx.conf'
+-)
++	'longer-timeout.patch')
+ sha256sums=('31f793bafda4a3259e9b1a4be0cb4dd1a70d856c2fd96db760cbbf64462f8cd6'
+-            '045dda10d978acedba43b071271731b2b63a8ad2743cdd6eda7dab2fd98a52ae')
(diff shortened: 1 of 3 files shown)


nominatim-ui/riscv64.patch
```diff
@@ -0,0 +1,28 @@
...
+ 
++  patch -p0 -i ../longer-timeout.patch
+   yarn install
+ }
+ 
+@@ -33,6 +35,7 @@ build() {
+ check() {
+   cd "${pkgname}-tags-v${pkgver}"
+ 
++  export PUPPETEER_EXECUTABLE_PATH="/usr/bin/chromium"
```
Xeonacid, line 25: Wondering why x86_64 don't need to manually set this. Some architecture specific config?

qyl27: Because of puppeteer, the library they used to launch chrome, it will download chrome from google which was not provided RISC-V prebuilt binaries.

qyl27: Please refer their docs: https://pptr.dev/guides/configuration  
In the first paragraph in this page.

Xeonacid: > Because of puppeteer, the library they used to launch chrome, it will download chrome from google which was not provided RISC-V prebuilt binaries.

Sounds like a bug for Arch Linux, as it declares `checkdepends=('chromium')` but chromium is not used really.  The `PUPPETEER_EXECUTABLE_PATH` setting might be used.

qyl27: This line was merged into archlinux.

qyl27:
This patch was merged into upstream in the next version of nominatim-ui.  
It could be merged for now and to be removed in the future.

Xeonacid:
CI is failing, there should be a checksum mismatch for the patch.

qyl27:
> CI is failing, there should be a checksum mismatch for the patch.

Fixed.

PR #3990, opened 2024-08-14
addpatch: cudd 3.0.0

Cryolitia:
[Upstream](https://github.com/ivmai/cudd) is actually an unofficial and inactive mirror, I would like to know whether should I still report it.


felixonmars:
Yeah it looks quite stale, unfortunately...

PR #4003, opened 2024-08-16
updpatch: jenkins 2.474-1

qyl27:
Fix timeout in riscv devices.


felixonmars:
Please try to make upstream aware of this too :)

PR #4062, opened 2024-09-06
qemu-user-blacklist.txt: add breezy

Cryolitia:
qemu-user does not handle prlimit64 with second argument set to RLIMIT_AS
correctly, failling `breezy/tests/blackbox/test_big_file.py`.

Reference: https://gitlab.com/qemu-project/qemu/-/blob/v9.1.0/linux-user/syscall.c?ref_type=tags#L13322-13351

### DIFF
--- a/qemu-user-blacklist.txt
+++ b/qemu-user-blacklist.txt
@@ -4,6 +4,7 @@ asio
 bat
 bear
 bmake
+breezy
 buildbot
 caps
 cargo-audit


hack3ric:
The patch is huge and only contains PGP key changes. Could you upstream it to Arch Linux instead?

felixonmars:
The exported key files are auto-generated and will be overridden on update, so I'm afraid it's not possible to upstream at the moment. I am also afraid that shipping the patch here is not a good idea for the same reason.

I'd suggest to send the right key to generic keyservers (at least ubuntu and the current "live" graph member like pgpkeys.eu). The tooling will try a keyserver first anyway :)

Cryolitia:
The key has been sent to keyserver.ubuntu.com / pgpkeys.eu, commit has been dropped. Still need this PR to add it to qemu-user-blacklist.

Do we need another PR to delete this patch: https://github.com/felixonmars/archriscv-packages/blob/bf14f89640fafa6b4727ecc894c61432a0c354ce/python-aiohttp-openmetrics/riscv64.patch

hack3ric:
Yes.

Cryolitia:
Opened at https://github.com/felixonmars/archriscv-packages/pull/4069

PR #4093, opened 2024-09-15
addpatch: qtile 0.28.1-1

ziyao233:
See https://gitlab.archlinux.org/archlinux/packaging/packages/qtile/-/merge_requests/1

### DIFF
--- a/qtile/riscv64.patch
+++ b/qtile/riscv64.patch
@@ -1,0 +1,16 @@
+diff --git a/PKGBUILD b/PKGBUILD
+index 09d5686..3325383 100644
+--- PKGBUILD
++++ PKGBUILD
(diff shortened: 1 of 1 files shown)


qtile/riscv64.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/PKGBUILD b/PKGBUILD
+index 09d5686..3325383 100644
+--- a/PKGBUILD
++++ b/PKGBUILD
```
Xeonacid, line 4:
```suggestion
--- PKGBUILD
+++ PKGBUILD
```

To adhere to our guidelines. :)

ziyao233: Thanks, fixed in the new HEAD

PR #4109, opened 2024-09-19
updpatch: uwsgi 2.0.26-3

qyl27:


uwsgi/riscv64.patch
```diff
@@ -14,46 +12,32 @@ index 8556394..c47cbbb 100644
...
    lua51
 -  mono
    openssl
    pam
-   pcre
+   pcre2
    perl
    php-embed
    php-legacy-embed
 -  pypy
```
Xeonacid, line 26: We have `pypy` now.

PR #4146, opened 2024-09-27
addpatch: colorhug-client 0.2.8

Cryolitia:
The [upstream](https://github.com/hughski/colorhug-client) is archived, can not create issue on it.


felixonmars:
Oops, I saw #4151 before this.

But these changes should be put in prepare() :)

qyl27:
I'm sorry for my fault, it won't happen again.

PR #4225, opened 2024-10-14
addpatch: cl-ppcre 2.1.2-1

kxxt:
Bootstrap cl-ppcre by building without cl-unicode. Tests are disabled because run-tests.lisp requires cl-unicode.


kxxt:
After that, we can build cl-unicode. And then we can finish the bootstraping by removing this patch and build cl-ppcre with the previously built two packages (`-I ../cl-unicode/cl-unicode-0.1.6.r11.g2790a6b-2-any.pkg.tar.zst -I ./cl-ppcre-2.1.2-1-any.pkg.tar.zst`)

felixonmars:
Ah, the patch is under a wrong folder :(

kxxt:
> Ah, the patch is under a wrong folder :(

Sorry about that, I forgot to change pwd when making this PR.

PR #4234, opened 2024-10-15
addpatch: health 0.95.0

Cryolitia:


health/riscv64.patch
```diff
@@ -0,0 +1,15 @@
...
+@@ -40,8 +40,11 @@ b2sums=(
+ prepare() {
+   cd Health
+ 
++  echo -e "\n[patch.crates-io]\nring = { git = 'https://github.com/felixonmars/ring', branch = '0.16.20' }" >> Cargo.toml
++  cargo update -p ring
++
+   CARGO_HOME="$srcdir/build/cargo-home" \
+-    cargo fetch --locked --target "$(rustc -vV | sed -n 's/host: //p')"
++    cargo fetch --locked
```
felixonmars, line 12: I guess this change is not needed?

Cryolitia: updated

PR #4275, opened 2024-11-13
updpatch: onnxruntime 1.19.2-2

felixonmars:
onnxruntime_USE_MPI was guarded with onnxruntime_USE_NCCL before, which is no longer the case now. The MPI path depends on AVX instrinsics for fp16, so we have to disable it for now.


felixonmars:
NOTE: This package requires RVV 1.0 on the build machine or if fails with signal 4 from protoc.

PR #4302, opened 2024-11-19
updpatch: chromium 131.0.6778.85-1

kxxt:
Arch switched to fetch the sources in prepare because chromium release tarball didn't get published. This PR instead uses the way we handle electron sources because patching gclient to ignore prebuilt dependencies become harder nowadays compared to electron 22 era.

Disable the download of pgo profiles because it is hard to patch gsutils which is downloaded and executed on the fly. (The pgo profiles are not used anyway due to chrome_pgo_phase=0).

Changes for Chromium 130
========================

- Replace std::hardware_destructive_interference_size with 64 because clang 18 doesn't support it for riscv64. (Not a problem for x86_64 Arch Linux). This can be dropped once we have clang 19.
- Add riscv-chromium-variations-130.patch to fix a crash(credits: https://github.com/revyos/chromium/blob/revyos/debian/patches/riscv64/fix-study-crash.patch)
- Drop upstreamed patch.
- Regenerate ffmpeg patch

Changes for Chromium 131
========================

- Add a patch for ffmpeg to fix linking problem regarding PIC. I have reported it to ffmpeg(https://trac.ffmpeg.org/ticket/11302) and it has been fixed in ffmpeg master.
- Regenerate ffmpeg patch


kxxt:
image

Tested on centiskorch

kxxt:
Patch is updated for 131.0.6778.85-1

PR #4303, opened 2024-11-19
addpatch: mongo-c-driver 1.29.0, qemu-user-blacklist: add mongo-c-driver

Cryolitia:
The test timeout will only randomly occurs on qemu-user, for example, it once costs 51.17 sec on Intel Ultra 9 185H. On both SG2042 and LicheePi 4A, it finished in just several seconds.


Cryolitia:
updated

PR #4321, opened 2024-11-24
addpatch: ruby 3.3.5-2

zinnjonas:
- test_x509req.rb needs adjustment for OpenSSL 3.3+ [https://github.com/ruby/openssl/pull/747/commits/c06fdeb0912221d9a2888369bbf9c10704af021e]
- test_https.rb needs to skip the test for OpenSSL 3.4 alike to OpenSSL 3.3

### DIFF
--- a/ruby/riscv64.patch
+++ b/ruby/riscv64.patch
@@ -1,0 +1,33 @@
+diff --git a/PKGBUILD b/PKGBUILD
+index f0d4d66..45d2636 100644
+--- a/PKGBUILD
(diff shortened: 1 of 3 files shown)


hack3ric:
Seems not RISC-V specific; does the tests fail in Arch x86 too? If it does please report to them first.

Besides, in terms of format:

- Please squash the changes in one commit
- Patch should start in `--- PKGBUILD`..., since we use `patch -p0`

PR #4366, opened 2024-12-16
updpatch: linux-tools 6.12-8

hack3ric:
- Remove intel-speed-select
- Fix `<asm/fence.h> not found` issue by cherry-picking https://github.com/torvalds/linux/commit/710fbca820c721cdd60fa8c5bbe9deb4c0788aae and set SRCARCH explicitly


hack3ric:
libpfm should be built first before linux-tools with this patch*

PR #4415, opened 2025-01-05
addpatch: chezmoi 2.57.0-1, qemu-user-blacklist: add chezmoi

Cryolitia:
Disable the test timeout, it's passed by args instead of hard encoding in the source, so may not report upstream.

It costs ~800s on `centiskorch` to finish, on qemu-user it will costs more than 1 hour (1 hour is the timeout set on upstream action CI) and I'm not able to wait it to finish, each step costs several minutes. But I believe it finally could be before the universe death hh

 Just be patient, we are using RISC-V.

### DIFF
--- a/chezmoi/riscv64.patch
+++ b/chezmoi/riscv64.patch
@@ -1,0 +1,11 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -46,7 +46,7 @@ build() {
+ check() {
+   cd "$pkgname-$pkgver"
+ 
+-  go test -v ./...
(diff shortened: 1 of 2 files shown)


felixonmars:
Hrm, if it passes on real hardware but is too slow for qemu-user, I'd prefer we add it to qemu-user-blacklist :)

(qemu-user specific problems are less likely to be upstreamed, or even mentioned to upstream at all. It's okay to disable it as an imperfect optimization.)

Cryolitia:
@Cryolitia 1

PR #4440, opened 2025-01-13
updpatch: libcaca 0.99.beta20-5

cybaol:


libcaca/riscv64.patch
```diff
@@ -1,14 +1,21 @@
...
  }
 +
-+source+=("$pkgname-fix-instruction-detection.patch::https://github.com/cacalabs/libcaca/pull/76.patch")
-+sha256sums+=('2d40a2c2654369e0583f9c26653b56d6b4306d76ad12c78478de489a8f91160b')
-+prepare() {
-+  cd $pkgname-$pkgver
-+  patch -Np1 -i ../$pkgname-fix-instruction-detection.patch
-+  autoreconf -fi
-+}
++source+=("libcaca-fix-instruction-detection.patch")
```
Xeonacid, line 20: Is the patch modified comparing with the upstream PR? If not, it is better to reference it by URL without storaging in the repo.

Xeonacid (approved):

PR #4441, opened 2025-01-14
updpatch: refind 0.14.2-1

cybaol:
https://sourceforge.net/p/refind/discussion/general/thread/acd89b2d43/

### DIFF
--- a/refind/riscv64.patch
+++ b/refind/riscv64.patch
@@ -1,20 +1,8 @@
+diff --git PKGBUILD PKGBUILD
+index 5ba99cb..926afba 100644
 --- PKGBUILD
 +++ PKGBUILD
-@@ -8,13 +8,17 @@ pkgdesc="An EFI boot manager"
- arch=(any)
- url="https://www.rodsbooks.com/refind/"
- makedepends=(bash dosfstools efibootmgr gnu-efi)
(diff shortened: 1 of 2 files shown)


refind/riscv64.patch
```diff
@@ -31,3 +19,22 @@
...
  
  package_refind() {
+@@ -67,8 +68,8 @@ package_refind() {
+   # manually
+   # efi binaries
+   install -vDm 644 refind/*.efi -t "$pkgdir/usr/share/$pkgname/"
+-  install -vDm 644 drivers_*/*.efi -t "$pkgdir/usr/share/refind/drivers_$_arch/"
+-  install -vDm 644 gptsync/*.efi -t "$pkgdir/usr/share/$pkgname/tools_$_arch/"
++  install -vDm 644 drivers_*/*.efi -t "$pkgdir/usr/share/refind/drivers_riscv64/"
++  install -vDm 644 gptsync/*.efi -t "$pkgdir/usr/share/$pkgname/tools_riscv64/"
```
Xeonacid, line 29: Better to change the `$_arch` variable

cybaol: You are right. But if do this, a diff of the hash value will be introduced in `riscv64.patch`, which will be broken after the next version upgrade.

Xeonacid: Would changing at the bottom line work? It should be sourced first and available when executing `package()`

PR #4456, opened 2025-01-19
updpatch: chromium-bsu 0.9.16.1-3

Cryolitia:
we need to firstlly run `gettextize -f && aclocal -I m4` to fix gettext infrastructure mismatch


chromium-bsu/riscv64.patch
```diff
@@ -1,11 +1,20 @@
...
  }
  
  build(){
+@@ -31,3 +34,7 @@ package() {
+ 
+   install -Dm644 COPYING "$pkgdir"/usr/share/licenses/chromium-bsu/LICENSE
+ }
++
++# run `gettextize -f && aclocal -I m4` in source
++source+=('gettext.patch')
```
Xeonacid, line 19: Can we run it when building instead of pre-generate?

Cryolitia: No, the script prevents itself being runned in automatic script by require an Return  key press in /dev/tty

Xeonacid: This should work:
```shell
  # Fake /dev/tty
  echo -e "\n" | script -qefc "gettextize -f" /dev/null
  aclocal -I m4
  autoreconf -fiv
```

Cryolitia: 图片
I tried, it doesn't work.

But why don't why just avoid using autoreconf. patch updated.

PR #4458, opened 2025-01-20
updpatch: nzbget 24.5-1

Cryolitia:
fix: `lto1: fatal error: target specific builtin not available`

### DIFF
--- a/nzbget/riscv64.patch
+++ b/nzbget/riscv64.patch
@@ -1,14 +1,8 @@
 --- PKGBUILD
 +++ PKGBUILD
-@@ -22,6 +22,11 @@ prepare() {
-   cd $pkgname-$pkgver
-   # https://github.com/nzbget/nzbget/pull/793
-   patch -Np1 -i ../nzbget_21.1_openssl3_update.patch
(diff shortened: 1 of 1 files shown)


nzbget/riscv64.patch
```diff
@@ -12,3 +18,8 @@
  }
  
  build() {
+   cd $pkgname-$pkgver
++  export CFLAGS+=" -fno-lto" CXXFLAGS+=" -fno-lto"
```
felixonmars, line 22: I guess we need only one of the export here or `!lto`? Do we really need both?

Cryolitia: I just copy it from https://github.com/felixonmars/archriscv-packages/blob/03effabe80dc00ef1a7c67f1df1ea2ce5cc95af7/snes9x/riscv64.patch#L10

I'll give it a try to just keep the one

cybaol: I only add `options+=('!lto')`, also build passed.

Cryolitia: @cybaol thx for telling me that! I have added you as the `Suggested-by`

PR #4469, opened 2025-01-21
addpatch: python-mitmproxy-rs 0.11.4-1

hack3ric:
Fix aya build: https://github.com/aya-rs/aya/pull/1139


tamird:
I don't know if you care but the way this patch is done can trivially compromise your supply chain since it downloads code from @hack3ric's fork which they can change at any time.

hack3ric:
Thanks for pointing it out. It is only a temporary patch, and will be removed once upstream adopted the fix. Neverless it should be done by pinning down the specific commit hash.

PR #4504, opened 2025-02-04
updpatch: transmission 4.0.6-9

cybaol:
Now we don't need to remove `libb64` from makedepends.
This is because makedepends changes frequently and thus cannot be patched.


hack3ric:
You can add `makedepends=(${makedepends[@]/libb64})` at the bottom of PKGBUILD.

PR #4439, opened 2025-01-13
addpatch: gitsign

cybaol:

PR #4520, opened 2025-02-11
addpatch: gitsign 0.10.2-1

Xeonacid:
Enable CGO. Upsteamed https://github.com/sigstore/gitsign/pull/631


Xeonacid:
Dup of #4439

PR #4526, opened 2025-02-12
addpatch: ot-simian

cybaol:


Xeonacid:
Please create an MR on Arch also, thanks.
Example: https://gitlab.archlinux.org/archlinux/packaging/packages/cargo-supply-chain/-/merge_requests/1

PR #4534, opened 2025-02-14
addpatch: aom

cybaol:
Disable `lto` to fix `lto1: fatal error: target specific builtin not available`


felixonmars:
Looks like this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=110812

PR #4539, opened 2025-02-15
addpatch: ginkgo-hpc 1.9.0-1

Xeonacid:
Disable CUDA.

Build after #4538 bumped.


Xeonacid:
Note: The Ci fails because the patch is based on the latest commit [db0f97de](https://gitlab.archlinux.org/archlinux/packaging/packages/ginkgo-hpc/-/commit/db0f97deec80b6737da988d944175cb6593d2220), which does re-format and has no functional change. It saves us the trouble of updating the patch the next time version bumps.

Please package from git `main` head. Thanks.

PR #4553, opened 2025-02-18
addpatch: naev

cybaol:
Same as #4363
But `export CFLAGS="$CFLAGS -DNO_fpu_control"` can not fix this.
So final solution: disable `fpu` related macro definition in `meson.build`.

### DIFF
--- a/naev/disable-riscv64-fpu-exception.patch
+++ b/naev/disable-riscv64-fpu-exception.patch
@@ -1,0 +1,13 @@
+diff -Naur naev-0.12.3/meson.build naev-0.12.3_/meson.build
+--- naev-0.12.3/meson.build	2025-02-01 10:29:56.000000000 +0800
++++ naev-0.12.3_/meson.build	2025-02-18 23:25:31.006265834 +0800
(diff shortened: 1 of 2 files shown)


naev/disable-riscv64-fpu-exception.patch
```diff
@@ -0,0 +1,13 @@
+diff -Naur naev-0.12.3/meson.build naev-0.12.3_/meson.build
+--- naev-0.12.3/meson.build	2025-02-01 10:29:56.000000000 +0800
++++ naev-0.12.3_/meson.build	2025-02-18 23:25:31.006265834 +0800
+@@ -275,9 +275,7 @@
+    summary('tracy', have_tracy, section: 'Debug', bool_yn: true )
+ 
+    # Standard library feature tests
+-   config_data.set10('HAVE_FEENABLEEXCEPT', cc.has_header_symbol('fenv.h', 'feenableexcept', prefix: '#define _GNU_SOURCE'))
```
kxxt, line 8: It would be better to fix this bad check and upstream the fix to naev rather than deleting this check.

PR #4568, opened 2025-02-24
addpatch: live-media 2025.01.17-1

gyx47:
Pulled the updated version


Xeonacid:
Unless it's a critical package, we generally don't update the package version here, but rather wait for Arch Linux to update.

Xeonacid:
IIUC, the upstream has removed tarball for old versions right? Then just waiting for Arch's upgrade should be fine. I think there is nothing that requires our action currently.
Thanks for your contribution!

gyx47:
should i upstream to Arch?

Xeonacid:
> should i upstream to Arch?

May not. They [discourages submitting Out-of-Date bug](https://wiki.archlinux.org/title/Bug_reporting_guidelines#Reasons_for_not_being_a_bug) and the package has already been marked as Out-of-Date at Archweb.

PR #4569, opened 2025-02-24
addpatch: vamp-aubio-plugins 0.5.1-6

kxxt:
- Add an else branch to set fallback value of local_vamp_lib to `/non-existent`. This variable is not used as we are using system vamp-plugin-sdk.
- Drop '-msse', '-msse2' compiler flags

Closes: https://archriscv.felixc.at/.status/log.htm?url=logs/vamp-aubio-plugins/vamp-aubio-plugins-0.5.1-6.log


felixonmars:
Upstream seems abandoned, but it would still be a good idea to at least try to activate them x)

PR #4596, opened 2025-03-08
addpatch: rbenv-1.3.2

gyx47:
dependence loss and noqumu.


rbenv/riscv64.patch
```diff
@@ -0,0 +1,22 @@
+diff --git PKGBUILD PKGBUILD
+index f16f5ce..588e843 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -12,7 +12,7 @@ pkgdesc="Manage your app's Ruby environment"
+ arch=('any')
+ url="https://github.com/rbenv/rbenv"
+ license=('MIT')
+-depends=('bash')
++depends=('bash' 'git')
```
Xeonacid, line 10: I guess `git` should be in checkdepends from the [failing log](https://archriscv.felixc.at/.status/log.htm?url=logs/rbenv/rbenv-1.3.2-1.log).

rbenv/riscv64.patch
```diff
@@ -0,0 +1,22 @@
...
+ checkdepends=('bats')
+ optdepends=('ruby-build: install Ruby versions')
+ install="rbenv.install"
+@@ -20,6 +20,8 @@ source=("$url/archive/v$pkgver/$pkgname-$pkgver.tar.gz")
+ sha256sums=('e2104f6472d7a8477409c46d4de39562b4d01899148a3dbed73c1d99a0b4bb2a')
+ 
+ check() {
++  export SHELL=/bin/bash
++  export RBENV_SHELL=bash
```
Xeonacid, line 19: Why do we need this? Is it the correct way to fix the failing tests?

gyx47: it turns out to be useless. QWQ

Xeonacid:
Once again, please modify your commit message and PR title in the correct format... Thanks for your effort.

rbenv/riscv64.patch
```diff
@@ -0,0 +1,13 @@
+diff --git PKGBUILD PKGBUILD
+index f16f5ce..f638331 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -13,7 +13,7 @@ arch=('any')
+ url="https://github.com/rbenv/rbenv"
+ license=('MIT')
+ depends=('bash')
+-checkdepends=('bats')
++checkdepends=('bats' 'git')
```
Xeonacid, line 10: Have you checked if the test passes on Arch Linux x86_64? If it fails, the patch should also be contributed to Arch.

gyx47: yes,i'll do it later.Sincerely sorry for the trouble I have caused to you.

Xeonacid:
> noqemu

Should be added to [qemu-user-blacklist.txt](https://github.com/felixonmars/archriscv-packages/blob/master/qemu-user-blacklist.txt)

Xeonacid (approved):

PR #4597, opened 2025-03-08
addpatch: ruby-pycall 1.5.2-4

gyx47:
fix 'Failure/Error: expect(out.chomp).to eq('true')' in riscv64 for Bundler/gem analysis.


Xeonacid:
Should this be fixed upstream (`ruby-pycall` itself) ?

PR #4598, opened 2025-03-08
qemu-user-blacklist.txt: add rbenv

gyx47:
add rbenv


Xeonacid:
It would be better to do this just in #4596. And the file should be alphabetically ordered.

PR #4627, opened 2025-03-27
addpatch: ruby-mini_portile2 2.8.8-1

saeziae:


felixonmars:
Would be great if this can be upstreamed :)

saeziae:
It happens in a packed "c-ares" inside upstream release, I don't know if we can let it change :)

PR #4633, opened 2025-03-29
updpatch: vault 1.19.0-1

Cryolitia:
Increase nodejs heap size to avoid OOM

### DIFF
--- a/sg2042-blacklist.txt
+++ b/sg2042-blacklist.txt
@@ -13,4 +13,3 @@ qt5-webengine
 qt6-webengine
 typescript
 grafana
-vault
--- a/vault/riscv64.patch
+++ b/vault/riscv64.patch
@@ -11,3 +11,11 @@
  }
  
  build() {
+@@ -48,6 +52,7 @@ build() {
+   export CGO_CXXFLAGS="${CXXFLAGS}"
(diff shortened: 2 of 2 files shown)


felixonmars:
Any idea why this is only needed on riscv64?

Cryolitia:
I could successfully build `vault` on `centiskorch` without the extra environment variable to increase heap limit. But we could still add it to make the build more stable.

kxxt:
> I could successfully build `vault` on `centiskorch` without the extra environment variable to increase heap limit. But we could still add it to make the build more stable.

Then it would be more interesting to know on what kind of machibe this package encountered heap OOM.

Historically this package cannot be built on sg2042 due to OOM and changing the variable cannot fix it: https://github.com/felixonmars/archriscv-packages/pull/3227

Please remove it from sg2042 blacklist if it could be built on sg2042 now. Thanks.

Cryolitia:
> Then it would be more interesting to know on what kind of machibe this package encountered heap OOM.

I guess this may be related to the number of threads, but there is no evidence yet, and it is difficult to stably reproduce an OOM build.

> Please remove it from sg2042 blacklist if it could be built on sg2042 now. Thanks.

Done

PR #4635, opened 2025-03-31
updpatch: gitlab 17.10.1-1

saeziae:
- resolve conflict
- patch nokogiri
- original `rbtrace` patch is modified because compiling should not be in the prepare() part

### DIFF
--- a/gitlab/riscv64.patch
+++ b/gitlab/riscv64.patch
@@ -1,6 +1,8 @@
+diff --git PKGBUILD PKGBUILD
+index 0010b42..a4b991b 100644
 --- PKGBUILD
 +++ PKGBUILD
-@@ -82,7 +82,18 @@ _etcdir=/etc/webapps/gitlab
+@@ -85,7 +85,14 @@ _etcdir=/etc/webapps/gitlab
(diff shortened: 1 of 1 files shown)


gitlab/riscv64.patch
```diff
@@ -11,30 +13,43 @@
...
++
+ 	cd gitlab-foss
+ 
+ 	# https://github.com/nodejs/node/issues/48444
+@@ -137,11 +149,22 @@ build() {
  	bundle config --local force_ruby_platform true # some native gems are not available for newer ruby
  	bundle config --local deployment true
  	bundle config --local without 'development test aws kerberos'
++	# override nokogiri build flags
++	bundle config --local build.nokogiri "--use-system-libraries --enable-shared --disable-static --with-cflags='-O2'"
```
Xeonacid, line 37: Wondering if this should be upstreamed to Arch Linux. 🤔

saeziae: Hello again. I don't know if upstreaming it to arch linux is up to the code of conduct, because according to investigation, x86 uses the prebuilt gem, so they are not compiling it natively. Adding this line to x86 build does no difference, the upside is that it will not affect the compiling behaviour, the downside, however, is that it is actually a meaningless line on x86 packing process.

PR #4637, opened 2025-03-31
addpatch: browserpass-firefox 3.9.0-2

gyx47:
Corrected integrity check


Xeonacid:
Note: The package has been flagged as out-of-date, so not reporting to Arch Linux should be fine.

PR #4712, opened 2025-04-30
addpatch: python-pyjsparser 2.7.1-14

gyx47:
After a simple test in the container, it is believed that there is no problem with the functionality, but the lack of js2py dependency on riscv64 makes the official test impossible. So I will skip the test for now and wait for js2py to be ported.upstreamed: [arch](https://gitlab.archlinux.org/archlinux/packaging/packages/python-pyjsparser/-/issues/1)

### DIFF
--- a/python-pyjsparser/riscv64.patch
+++ b/python-pyjsparser/riscv64.patch
@@ -1,0 +1,23 @@
+diff --git PKGBUILD PKGBUILD
+index 1f52c4b..601a63b 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -10,7 +10,7 @@ license=('MIT')
+ arch=('any')
+ depends=('python')
+ makedepends=('python-setuptools')
+-checkdepends=('python-js2py' 'python-pytest')
++checkdepends=('python-pytest')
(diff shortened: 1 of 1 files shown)


python-pyjsparser/riscv64.patch
```diff
@@ -0,0 +1,23 @@
+diff --git PKGBUILD PKGBUILD
+index 1f52c4b..601a63b 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -10,7 +10,7 @@ license=('MIT')
+ arch=('any')
+ depends=('python')
+ makedepends=('python-setuptools')
+-checkdepends=('python-js2py' 'python-pytest')
++checkdepends=( 'python-pytest')
```
Xeonacid, line 10:
```suggestion
+checkdepends=('python-pytest')
```

Xeonacid (changes requested):
Please drop the `riscv64.patch:Zone.Identifier` file. Thanks.

Xeonacid:
It seems that the `python-js2py` package has been dropped in Arch. This should also be reported to Arch due to a missing checkdepends.

PR #4713, opened 2025-04-30
addpatch: telepathy-salut 0.8.1-10

gyx47:
The config.guess / config.sub included in the package is too old and does not recognize riscv64.

### DIFF
--- a/telepathy-salut/riscv64.patch
+++ b/telepathy-salut/riscv64.patch
@@ -1,0 +1,16 @@
+diff --git PKGBUILD PKGBUILD
+index dca2c3d..e72fea1 100644
+--- PKGBUILD
++++ PKGBUILD
+@@ -31,7 +31,10 @@ prepare() {
+ 
+ build() {
+     cd "$pkgname-$pkgver"
+-    ./configure --prefix=/usr \
++    ./configure \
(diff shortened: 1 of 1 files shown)


Xeonacid:
Could you please check if an `autoreconf -fiv` solves the problem? If true, it can be upstream to Arch per the [Arch-dev-public mailing list](https://lists.archlinux.org/archives/list/arch-dev-public@lists.archlinux.org/message/WN7T54L266NKN33ERC4PHOTYOQA7OYK5/).

gyx47:
it works, do I need to submit PR again?

gyx47:
autoreconf -fiv can solve the problem of the old config.guess/sub, but it regenerates the entire Autotools stack, which requires gtk-doc. After I added the gtk-doc package and ran it again, I found that the generated makefile still failed when running.
```
/build/telepathy-salut/src/telepathy-salut-0.8.1/lib/ext/wocky/wocky/wocky-xmpp-error-enumtypes.h:12:9: error: macro names must be identifiers
   12 | #define /BUILD/TELEPATHY_TYPE_XMPP_ERROR_TYPE (wocky_xmpp_error_type_get_type())
      |         ^

```

felixonmars:
Would still be good to report it to upstream :)

PR #4772, opened 2025-05-26
addpatch: mattermost-desktop 5.12.0-1

hack3ric:
- Use forked electron-builder and other npm packages
- Remove all explicit x64 mentions
- Manually call electron-builder so it uses system electron only (adopted from bitwarden)

### DIFF
--- a/mattermost-desktop/riscv64.patch
+++ b/mattermost-desktop/riscv64.patch
@@ -1,0 +1,57 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -49,23 +49,51 @@ prepare() {
+ 			--arg electronVersion "$_electronVersion" \
+ 			package.json |
+ 		sponge package.json
++	jq '	.devDependencies."electron-builder"="npm:@riscv-forks/electron-builder@24.13.3" |
++		.overrides."app-builder-lib"="npm:@riscv-forks/app-builder-lib@24.13.3" |
++		.overrides."builder-util"="npm:@riscv-forks/builder-util@24.13.1"' package.json |
++		sponge package.json
+ 	sed -i 's/ && cross-env CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ electron-builder --linux tar.gz --arm64 --publish=never//g' package.json
++	sed -i 's/ --x64//g' package.json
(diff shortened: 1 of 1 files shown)


mattermost-desktop/riscv64.patch
```diff
@@ -0,0 +1,57 @@
...
+ 	sed -i 's/ && cross-env CC=aarch64-linux-gnu-gcc CXX=aarch64-linux-gnu-g++ electron-builder --linux tar.gz --arm64 --publish=never//g' package.json
++	sed -i 's/ --x64//g' package.json
++	# Ugly workaround for https://github.com/npm/cli/issues/5443 https://github.com/npm/cli/issues/7660
++	rm -f package-lock.json
++	# Remove unused postinstall script (electron-rebuild)
++	sed -i '/"postinstall":/d' package.json
++	export ELECTRON_SKIP_BINARY_DOWNLOAD=1
+ 	npm $_npmargs install
++
++	local _builder_bin=node_modules/app-builder-bin/linux/riscv64
```
kxxt, line 20: You can use `@riscv-forks/builder-util=25.1.7` `@riscv-forks/app-builder-lib=25.1.8` `@riscv-forks/electron-builder=25.1.8` that depends on a newer app-builder supporting riscv64 to avoid vendoring it here. See also https://github.com/felixonmars/archriscv-packages/pull/4773

hack3ric: Seems they are still depending on 24.13.x, and upgrading it to 25.1.x breaks build:

```
 • electron-builder  version=25.1.8 os=6.6.46-pioneer
  • loaded configuration  file=/build/mattermost-desktop/src/desktop-5.12.0/electron-builder.json
  • executing @electron/rebuild  electronVersion=35.4.0 arch=riscv64 buildFromSource=false appDir=./
  • installing native dependencies  arch=riscv64
  ⨯ ENOENT: no such file or directory, stat '/build/mattermost-desktop/src/desktop-5.12.0/node_modules/@mattermost/desktop-api'  failedTask=build stackTrace=Error: ENOENT: no such file or directory, stat '/build/mattermost-desktop/src/desktop-5.12.0/node_modules/@mattermost/desktop-api'
```

Let's keep it as is now, and recheck afterwards.

PR #4779, opened 2025-05-30
updpatch: chromium 137.0.7151.40-1

kxxt:
- Packaging changes:
  - `-march=rv64gc` in CFLAGS overrides some objects from xnnpack that should be complied with rv64gcv. Xnnpack has runtime detection for RVV. Currently I cannot think up a better solution so drop `-march` in CFLAGS.
- RISC-V patchset changes:
  - Drop merged patches
  - Regenerate ffmpeg patch
  - Refresh sandbox patch
  - Enable cpuinfo build for riscv64, upstreamed: https://crrev.com/c/6597694
  - Backport patch for riscv64 xnnpack build: https://crrev.com/c/6597834

### DIFF
--- a/chromium/riscv64.patch
+++ b/chromium/riscv64.patch
@@ -9,7 +9,7 @@
  optdepends=('pipewire: WebRTC desktop sharing under Wayland'
              'kdialog: support for native dialogs in Plasma'
              'gtk4: for --gtk-version=4 (GTK4 IME might work better on Wayland)'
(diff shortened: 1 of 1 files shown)


kxxt:
> Tested on centiskorch. It crashed. I will debug it later

Looks like it is because previously I tested mainline chromium, which upgrades the database, causing incompatibility with chromium 137. The crash is solved after clearing `~/.{config,cache}/chromium`

PR #4817, opened 2025-06-30
addpatch: mate-calc 1.28.0-3

gyx47:
"undefined macro" because gettext is not installed on the system in makedepends, and aclocal cannot find its macro directory.upstreamed:[arch](https://gitlab.archlinux.org/archlinux/packaging/packages/mate-calc/-/issues/1)


Xeonacid:
> "undefined macro" because gettext is not installed on the system in makedepends

That's not true. gettext is always installed as it is a dependency of base-devel.

IIRC the error is caused by a behavior change in the recent gettext.

Some details at https://gitlab.archlinux.org/archlinux/packaging/packages/dovecot/-/issues/4#note_268535

PR #4843, opened 2025-08-02
updpatch: librustls 0.15.0-2

kxxt:
Create a cmake wrapper to add `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to avoid build failure with cmake 4.x.

### DIFF
--- a/librustls/riscv64.patch
+++ b/librustls/riscv64.patch
@@ -7,5 +7,35 @@
 +  cmake
 +  clang
  )
- provides=('librustls.so')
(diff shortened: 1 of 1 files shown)


felixonmars:
Would still be great to upstream this in some way :)

kxxt:
> Would still be great to upstream this in some way :)

I think soon we will no longer need it when upstream release a new version with updated aws-lc-sys  that uses prebuilt bindings instead of relying on cmake.

PR #4929, opened 2025-09-16
addpatch: basis-universal 1.60-1

gyx47:
SSE isn't available in riscv64


SpriteOvO (approved):
Nice work! Could you upstream this patch to Arch repo? Similar  to [this commit](https://gitlab.archlinux.org/archlinux/packaging/packages/cern-vdt/-/commit/e818a1fcd75eda1db1558832c0de28153417cb8d), that would be more helpful to our long term goal. Thanks.

PR #4931, opened 2025-09-18
addpatch: chewing-editor 0.1.2-1

gyx47:
On riscv64, Valgrind fails to start after failing to redirect dynamic linker symbols.  [upstreamed](https://bugs.kde.org/show_bug.cgi?id=510193)


gyx47:
https://bbs.archlinux.org/viewtopic.php?id=276422 is about the same problem.i'll have a try

gyx47:
[upstreamed](https://bugs.kde.org/show_bug.cgi?id=510193)

PR #4938, opened 2025-09-19
addpatch: calcmysky 0.3.5-1

gyx47:
The package can be packaged normally in the container, but the parallel compilation is stuck under qemu-user simulation.  run it with a single thread.


felixonmars:
With parallel enabled, it appears to deadlock on real hardware as well. Any idea why this doesn't happen on x86_64?

PR #4978, opened 2025-10-17
updpatch: chromium 141.0.7390.122-1

kxxt:
- Disable `_manual_clone` since the chromium source tarball has been released.
- Drop upsteamed modification to compiler-rt-adjust-paths.patch


kxxt:
Rebased patch on new release `141.0.7390.122-1`.

PR #4983, opened 2025-10-25
addpatch: dtkcore 5.7.23.-1

gyx47:
riscv64 build() crashes when running LeakSanitizer."LeakSanitizer does not work under ptrace"


felixonmars:
The error you got seems to be a limitation under qemu-user? The original error log suggests actual memory leak I think.

PR #4996, opened 2025-11-12
updpatch: java8-openjdk 8.472.u08-1

kxxt:
- Refresh patch.
- This needs to be built from main branch as it contains fixes for the build.


kxxt:
> It seems there are more issues now:
> 
> ```
> checking for X11/Intrinsic.h... no
> configure: error: Could not find all X11 headers (shape.h Xrender.h XTest.h Intrinsic.h).
> No configurations found for /build/java8-openjdk/src/jdk8u-jdk8u472-b08/! Please run configure to create a configuration.
> Makefile:55: *** Cannot continue.  Stop.
> ==> ERROR: A failure occurred in build().
> ```

I have opened a MR in Arch Linux gitlab: https://gitlab.archlinux.org/archlinux/packaging/packages/java8-openjdk/-/merge_requests/5

PR #5055, opened 2025-12-29
addpatch: podman-desktop 1.23.1-1

gyx47:
Vite 4.4 experimentally supports Lightning CSS.But there is currently no official LightningCSS binary for RISC-V; references to the Tailwind Vite plugin have been removed. Also, missing dependencies electron38


Xeonacid (commented):
> Also, the electron38 package failed to load.

Could you please leave some log here? So anyone who comes across here can know what is not working. Thanks!

PR #5189, opened 2026-03-31
updpatch: dart 3.11.4-1

kxxt:
- Add a patch to disable `-Werror` because of:

        /usr/include/c++/15.2.1/bits/shared_ptr_base.h:181:35: error: ‘atomic_thread_fence’ is not supported with ‘-fsanitize=thread’ [-Werror=tsan]

- Refresh patch
- This depends on tsan support in GCC: #5188


felixonmars:
Oops. I somehow missed this PR and introduced #5351 instead. Sorry!

PR #5204, opened 2026-04-03
updpatch: opensearch-sql-plugin 3.5.0.0-1

kxxt:
Use system gradle. Bundled gradle is missing gradle/native-platform@bc65f1d


felixonmars:
Seems not needed anymore. Still thanks!

PR #5214, opened 2026-04-08
addpatch: rage-encryption 0.11.1-3

Xeonacid:
Backport https://github.com/str4d/rage/commit/6b097057f65aae50860ef1f0417df2bf4a26cbde

`--target host-tuple` in `cargo fetch` is removed because `cargo check` requires downloading a dependency which is not compiled for the current arch...


Xeonacid:
Note: CI fails because the patch is based on the `HEAD` of the `main` branch to avoid patch failure in the future.

felixonmars:
No longer needed now. Thanks still!

PR #5327, opened 2026-05-06
Bootstrap GCC Ada

kxxt:
This PR bootstraps `gcc-ada` package. Some steps could run in parallel as shown in the following chart. 

```mermaid
flowchart TD
S1[Step1: <br>Build gcc 16.1.1 natively with gcc 15 on riscv64] --> S4
S2[Step2: <br>Build gcc cross-compiler with Ada support] --> S3[Step3: <br>Cross target riscv64 Ada compiler from x86_64]
S3 --> S4[Step4: <br>Build gcc 16.1.1 natively on riscv64 with Ada from Step3]
S4 --> S5[Step5: <br>Build gcc 16.1.1 with itself]
```

## Step1: Build gcc 16.1.1 natively with gcc 15 on riscv64

This is PR https://github.com/felixonmars/archriscv-packages/pull/5326

We need to update to 16.1.1 first as the bootstrap process here uses gcc 16.1.1.
This ensures that we are using the same version of the compiler across the whole bootstrap process.

## Step2: Build gcc cross compiler with Ada support

On Arch Linux x86_64, use gcc `16.1.1+r12+g301eb08fa2c5-1` to build `riscv64-linux-gnu-gcc` package from `gcc-ada-bootstrap/riscv64-linux-gnu-gcc` directory.

We get an Ada cross-compiler that could cross compile to riscv64 from this step.

(build=x86_64-linux-gnu, host=x86_64-linux-gnu, target=riscv64-linux-gnu)

## Step3: Cross target riscv64 Ada compiler from x86_64

On Arch Linux x86_64, use gcc `16.1.1+r12+g301eb08fa2c5-1` and `riscv64-linux-gnu-gcc` package built in previous step to cross build a native riscv64 Ada compiler.

We will get a `riscv64-unknown-linux-gnu-target-gcc-ada` package from this step.
This package uses `/opt/gcc-ada` prefix to avoid conflict with our gcc package.

```bash
extra-x86_64-build  -- -I ~/p/riscv64-linux-gnu-gcc/riscv64-linux-gnu-gcc-16.1.1+r12+g301eb08fa2c5-2-x86_64.pkg.tar.zst
```

(build=x86_64-linux-gnu, host=riscv64-linux-gnu, target=riscv64-linux-gnu)

## Step4: Build gcc 16.1.1 natively on riscv64 with Ada from Step3

Build gcc 16.1.1 with the gcc compiler from Step1 and the ada compiler from step3.
[...]

PR #5329, opened 2026-05-12
updpatch: gcc 16.1.1+r12+g301eb08fa2c5-1

kxxt:
This is the final step of the GCC Ada bootstrap in #5327.

Perhaps it's time to upstream this patch to Arch.


kxxt:
The patch test failed because it failed to clone the gcc git repo :(

PR #5342, opened 2026-05-16
addpatch: gmp 6.3.0-3

felixonmars:
nettle 4.0 check() fails on riscv64 in the Valgrind side-channel wrappers for RSA/ECC/EdDSA. The common frame is GMP RISC-V __gmpn_sec_tabselect, where the inner loop branch compares values derived from the secret table index, which Valgrind treats as secret-dependent control flow.

Patch GMP instead of disabling nettle tests: keep the masked table selection using which, but drive loop termination from the public nents count so the function still scans every table entry while branch operands stay public.

Use @dramforever's improved version, which allocates k to s1 directly instead of adding a separate cnt alias.

After this GMP fix, nettle and nettle3 still have one remaining Valgrind side-channel failure in sc-ecdsa-sign, now inside nettle's own secp256r1 modq reduction. That issue is harder to fix properly at the moment, so keep this patch scoped to the GMP sec_tabselect problem.

### DIFF
--- a/gmp/gmp-riscv64-sec-tabselect-public-loop.patch
+++ b/gmp/gmp-riscv64-sec-tabselect-public-loop.patch
@@ -1,0 +1,109 @@
+From: Felix Yan <felixonmars@archlinux.org>
+Subject: [PATCH] mpn/riscv64: keep sec_tabselect loop control public
+
+
+The current RISC-V mpn_sec_tabselect implementation derives the inner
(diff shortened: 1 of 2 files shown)


felixonmars:
GMP patch sent to devel mailing list, and the remaining nettle issue reported at https://git.lysator.liu.se/nettle/nettle/-/work_items/18

PR #5401, opened 2026-06-22
updpatch: talkfilters 2.3.8-8: refresh config.guess via autoreconf

felixonmars:
The riscv64 build fails because the bundled 2007 config.guess cannot identify riscv64. Run autoreconf -fiv in prepare() to refresh config.guess and config.sub before configure.

Arch 2.3.8-8 already passes CFLAGS=-Wno-format-security, so drop the old local wrap.c format-security patch.

Failure log: https://archriscv.felixc.at/.status/logs/talkfilters/talkfilters-2.3.8-8.log

Arch commit: https://gitlab.archlinux.org/archlinux/packaging/packages/talkfilters/-/commit/adb2946167e1e18c7131ee7746e16078793d96af


felixonmars:
https://gitlab.archlinux.org/archlinux/packaging/packages/talkfilters/-/merge_requests/2

PR #5576, opened 2026-09-05
updpatch: python-llvmlite 0.49.0-1

felixonmars:
Enable MCSaveTempLabels for RISC-V. LLVM emits distinct temporary symbols as ".L0 ", while RuntimeDyld resolves relocations by name. The collision corrupts jump tables, producing wrong results and Numba 0.67.0 segfaults. Add regressions for direct MCJIT execution and emitted-object loading.

Remove obsolete MCJIT, global-constructor and object-loading exclusions; the existing cases pass with LLVM 22. OrcLLJIT is already skipped upstream.

Replace the optsize/minsize skip with an explicit vector-capable x86_64 target for this IR-only test. Keep its vectorization assertions without assuming the native default CPU has vector registers. https://github.com/numba/llvmlite/issues/1449

### DIFF
--- a/python-llvmlite/fix-optsize-minsize-test.patch
+++ b/python-llvmlite/fix-optsize-minsize-test.patch
@@ -1,0 +1,44 @@
+--- a/llvmlite/tests/test_binding.py
++++ b/llvmlite/tests/test_binding.py
+@@ -2810,10 +2810,24 @@
+         self.assertIn("alloca", optimized_asm_optnone)
+ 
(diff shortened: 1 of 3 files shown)


mr-c:
Have these patches been proposed to the llvmlite people?

felixonmars:
> Have these patches been proposed to the llvmlite people?

Not yet, we need to evaluate how to address that MCJIT issue properly.

PR #5599, opened 2026-09-06
addpatch: deja-dup 50.2-1

felixonmars:
Allow a null destination in RecursiveOp.raise_error. RecursiveDelete has no destination, so forwarding a child error otherwise aborts with "__lambda11_: assertion 'd != NULL' failed".

The custom-tool-wrapper-bad test hits this path when cleanup races with GVfs renaming a temporary metadata file and unlink returns ENOENT. The same guard exists on x86_64; fix the error-reporting contract rather than blacklisting qemu-user or skipping the test.


felixonmars:
Patch submitted at https://gitlab.gnome.org/World/deja-dup/-/merge_requests/120

PR #5615, opened 2026-09-06
addpatch: ruby-ethon 0.18.0-1

felixonmars:
Gate the NTLM specs on the libcurl VERSION_NTLM feature bit. The check fails with HTTP 401 instead of 200 because the Arch curl 8.21.0 build lacks NTLM, which curl now disables by default. Keep the specs enabled when NTLM is available and leave runtime code unchanged.

The official package .BUILDINFO records an x86_64 build on 2026-03-14 with curl 8.19.0, before the 2026-03-21 default change. The difference is dependency age, not CPU architecture.

Curl change: https://github.com/curl/curl/commit/cc0c83c5f853c41fca6880c5dcd6745da9353434


felixonmars:
Upstreamed at https://github.com/typhoeus/ethon/pull/275

PR #5617, opened 2026-09-06
addpatch: criterion 2.4.3-1

felixonmars:
BoxFort writes an 18-byte RISC-V trampoline into main's 16-byte PLT entry, corrupting the next entry and causing SIGILL in syslog-ng tests on both native and QEMU builders. Use compressed load and jump instructions to fit the trampoline in 16 bytes.

Keep criterion and syslog-ng off Sv39 builders: BoxFort's inheritable arena requires addresses above 32 TiB, outside the Sv39 user address space. Criterion's tests abort with "Could not initialize inheritable arena: Cannot allocate memory" on baxcalibur.

Upstream: https://github.com/Snaipe/BoxFort/issues/44

### DIFF
--- a/criterion/boxfort-riscv64-plt.patch
+++ b/criterion/boxfort-riscv64-plt.patch
@@ -1,0 +1,27 @@
+--- a/src/asm/trampoline-riscv64.S
++++ b/src/asm/trampoline-riscv64.S
+@@ -29,17 +29,15 @@
+ #endif
+ .globl MANGLE(bxfi_trampoline)
+ MANGLE(bxfi_trampoline):
++#ifdef __riscv_compressed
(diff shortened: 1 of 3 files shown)


felixonmars:
Patch submitted at https://github.com/Snaipe/BoxFort/pull/57

MrAnno:
@felixonmars Do you think it would be worth reviving https://github.com/Snaipe/BoxFort/pull/49 as well?

(I'm completely unfamiliar with the risc world.)

felixonmars:
I think so, but I'm also unsure about the runtime detection bit. Will look into that :)

MrAnno:
Thanks :) Let me know if I can help in any way; I'll have some free time in the evening.

PR #5630, opened 2026-09-07
addpatch: haskell-primitive 0.9.1.0-209

felixonmars:
Convert narrow signed fill arguments to their unsigned equivalents before calling the C helpers. The mismatched FFI declarations sign-extend negative Int16 values on riscv64, making a fill with -2 produce [-2,-1,-1,-1] and causing 13 test failures. Apply the conversion to arrays and pointers.


felixonmars:
Patch submitted at https://github.com/haskell/primitive/pull/445

PR #5658, opened 2026-09-08
addpatch: ocicl 2.17.0-1

felixonmars:
Use ELT instead of AREF for the list-valued output in bundled IDNA's punycode-decode. The riscv64 build stops at a COMPILE-FILE-ERROR for idna/decode; the same failure is reported upstream with a LIST versus VECTOR type warning at these accesses.

https://github.com/antifuchs/idna/issues/2


felixonmars:
Patch submitted at https://github.com/antifuchs/idna/pull/3

PR #5672, opened 2026-09-08
addpatch: python-logbook 1.10.1-1

felixonmars:
The TCP syslog tests use listen(0), causing connect() to time out on kernels without SYN cookies. Use a backlog of one so the test's connection can complete before accept() is called.


felixonmars:
Patch submitted at https://github.com/getlogbook/logbook/pull/636

PR #5729, opened 2026-09-09
addpatch: python-quart-trio 0.12.0-2

felixonmars:
Adapt to Quart 0.23, which removes RequestContext and WebsocketContext in favor of AppContext. urllib3 2.7.0 fails during test collection with "cannot import name RequestContext from quart.ctx" on Quart 0.23.1. Pass the unified context through dispatch, finalization, exception handling and background-task callbacks.

Provide Trio-backed queues for the new HTTP body and WebSocket buffer APIs, including completion and disconnect handling. Reuse Quart-backed WebSocket construction and update the affected tests, adding streaming and disconnect coverage.

Remove the test-client imports of HTTPDisconnectError and WebsocketDisconnectError, which Quart 0.23 also removed and which block collection of all five test modules. Drop obsolete HTTP disconnect send-event handling and use ClientDisconnectedError for WebSocket closes, preserving the close-code argument.

Require Quart >=0.23 and Python >=3.13 in project metadata. Rebuild python-quart-trio before retrying urllib3; no urllib3 test exclusions are needed for this dependency failure.

### DIFF
--- a/python-quart-trio/quart-0.23.patch
+++ b/python-quart-trio/quart-0.23.patch
@@ -1,0 +1,492 @@
+--- a/pyproject.toml
++++ b/pyproject.toml
+@@ -28,10 +28,10 @@
+ dependencies = [
+     "exceptiongroup >= 1.1.0; python_version < '3.11'",
+     "hypercorn[trio] >= 0.12.0",
+-    "quart >= 0.19",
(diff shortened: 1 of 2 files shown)


felixonmars:
Patch submitted at https://github.com/pgjones/quart-trio/pull/23

PR #5780, opened 2026-09-11
addpatch: python-warlock 2.1.0-2

felixonmars:
Allow flit_core 4 in the build-system requirement. The riscv64 build fails with "ERROR Missing dependencies: flit_core~=3.2" despite having flit_core 4.0.2 installed. Keep the minimum version and cap at <5.

No upstream fix was found. Warlock 2.1.0 uses PEP 621 metadata, which Flit 4 supports.
https://github.com/bcwaldon/warlock/blob/2.1.0/pyproject.toml https://github.com/pypa/flit/blob/4.0.2/doc/history.rst

### DIFF
--- a/python-warlock/riscv64.patch
+++ b/python-warlock/riscv64.patch
@@ -1,0 +1,11 @@
+--- PKGBUILD
++++ PKGBUILD
+@@ -15,6 +15,8 @@
+ 
+ prepare() {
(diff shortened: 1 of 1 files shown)


felixonmars:
Submitted at https://github.com/bcwaldon/warlock/pull/71

PR #5844, opened 2026-09-13
addpatch: mypaint 2.0.1-8

felixonmars:
Replace MyPaint's stale PyString_FromString call with PyUnicode_FromString. SWIG 4.5 removed its Python 2 compatibility aliases, exposing the undefined C API name while generating Python 3 bindings.

Fedora carries the identical fix. No MyPaint upstream fix was found; both v2.0.1 and current master retain the stale call.

https://archriscv.felixc.at/.status/logs/mypaint/mypaint-2.0.1-8.log

https://github.com/swig/swig/commit/79f7a2b7cb7ddc256eba09c8770133148025171d

https://src.fedoraproject.org/rpms/mypaint/c/dbb6bb41ad9401799bd633691d52eba825694e4d

https://github.com/mypaint/mypaint/blob/v2.0.1/lib/tiledsurface.hpp#L131-L136

https://gitlab.archlinux.org/archlinux/packaging/packages/mypaint/-/blob/2.0.1-8/PKGBUILD


felixonmars:
Patch submitted at https://github.com/mypaint/mypaint/pull/1331

